PR #1, opened 2018-01-31
Focus frames beneath a node on dblclick

pushrax:
As discussed: double clicking a node will position that node at the top left of the view, and resize horizontally to fit that node.

screen shot 2018-01-30 at 23 38 19

becomes

screen shot 2018-01-30 at 23 38 34


jlfwong:
Yaaaaaaay first contribution ✨

PR #13, opened 2018-04-09
WIP: Add call-graph view

### LINKED ISSUE #11: Add call tree view
This should be helpful in cases where an expensive function shows up in many different stacks (e.g. if you had an expensive function used at various recursive depths in a compiler).

Inspired by the tree view in KCacheGrind: https://kcachegrind.github.io/html/Shot3.html

jlfwong:
Fixes #11


alangpierce:
@jlfwong screenshot please? 😄

I'd be happy to take a look at the code if you want another set of eyes.

jlfwong:
@alangpierce This is pretty far from done :P The code in here does layout of a DAG, but it isn't even wired up to profiling data yet, and it doesn't have pan/zoom, etc. Probably another couple of weeks before this is going to land.

jlfwong:
I'm abandoning this because I no longer think this is a good idea. It's really difficult to identify a useful subgraph to display without some additional filtering controls, and I think a modified bottoms-up view is probably more useful.

PR #19, opened 2018-04-15
Add profileURL and title hash parameters

alangpierce:
On init, we check the hash fragment for these parameters and load the URL. We
always show a loading state in that case rather than the landing screen.

When determining the title, an explicitly-specified title takes precedence,
otherwise we use the filename.

I also added an error state, currently only used for my new code, but possibly
there could be a more robust or widespread error handling approach in the
future.

### DIFF
--- a/application.tsx
+++ b/application.tsx
@@ -12,6 +12,7 @@ import {Profile, Frame} from './profile'
 import {Flamechart} from './flamechart'
 import {FlamechartView} from './flamechart-view'
 import {FontFamily, FontSize, Colors} from './style'
+import getHashParams, {HashParams} from './hash-params'
 
 declare function require(x: string): any
 const exampleProfileURL = require('./sample/perf-vertx-stacks-01-collapsed-all.txt')
@@ -29,6 +30,7 @@ interface ApplicationState {
   sortedFlamechartRenderer: FlamechartRenderer | null
   sortOrder: SortOrder
   loading: boolean
+  error: boolean
 }
 
 interface ToolbarProps extends ApplicationState {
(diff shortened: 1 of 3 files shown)


application.tsx
```diff
@@ -430,6 +454,15 @@ export class Application extends ReloadableComponent<{}, ApplicationState> {
     )
   }
 
+  renderError() {
+    return (
+      <div className={css(style.error)}>
+        <div>😿 Something went wrong.</div>
```
jlfwong: Nice :P

hash-params.ts
```diff
@@ -0,0 +1,28 @@
+export interface HashParams {
+  profileURL?: string
+  title?: string
+}
+
+export default function getHashParams(): HashParams {
```
jlfwong: I haven't really been using default exports for anything anywhere, and IIRC, it makes mocking things kind of a pain. Would you might switching this to just a regular exported function?

hash-params.ts
```diff
@@ -0,0 +1,28 @@
...
+  try {
+    const hashContents = window.location.hash
+    if (!hashContents.startsWith('#')) {
+      return {}
+    }
+    const components = hashContents.substr(1).split('&')
+    const result: HashParams = {}
+    for (const component of components) {
+      let [key, value] = component.split('=')
```
jlfwong: `const`?

application.tsx
```diff
@@ -343,8 +352,23 @@ export class Application extends ReloadableComponent<{}, ApplicationState> {
     }
   }
 
-  componentDidMount() {
+  async componentDidMount() {
```
jlfwong: It's a little weird having this function which is part of React's APIs return a promise. Can you extract the code that does the hash loading into an async method and just invoke it from `componentDidMount`?

alangpierce: Fair enough. I've done this in my own code without anything going wrong, but agreed it's a little fishy.

hash-params.ts
alangpierce: Oops, thanks.

hash-params.ts
alangpierce: Sure. Not sure about the details of the mocking you're talking about, but at least with the normal commonjs transform, it is possible to mock default exports, I'm pretty sure.

PR #59, opened 2018-06-13
Introduce a simple table view

jlfwong:
This is a pretty barebones table view. Eventually I'd like this to provide some better way of identifying problem areas by seeing flamegraphs of all callers and all callees, but this is better than nothing.

image


jlfwong:
Need to fix layout in Firefox before I can land this

PR #83, opened 2018-07-01
Specify file format for speedscope

### LINKED ISSUE #65: Document, export, and re-import a speedscope json file format
Since this app can read many trace formats, it would be pretty cool if we could have an export function into some sort of common JSON format.  That way we can use it to convert things like proprietary instruments traces into open chrome traces so we can use it with other tools.

jlfwong:
This defines a JSON-based file format for speedscope.

The motivation for is primarily two things:
1. To enable others to write tools to output profiles which can be read by speedscope
2. To enable others to write tools to handle the output of speedscope, leveraging the variety of importers that speedscope supports

Fixes #65


file-format-spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  // List of all call frames
+  frames: SerializedFrame[]
+
+  // List of nodes in the call tree
+  nodes: SerializedNode[]
+
+  // List of indices into nodes, with -1 indicating that the call-stack
+  // was empty at the time of the sample
+  samples: number[]
```
uber-mahyar2: Would it be possible to have an empty samples array for trace sources that are not using sampling?

file-format-spec.ts
```diff
@@ -0,0 +1,58 @@
+// This file contains types which specify the speedscope file format.
+
+export interface SerializedFrame {
+  name: string
+  file?: string
+  line?: number
+  col?: number
+}
+
+export interface SerializedNode {
```
uber-mahyar2: would it be possible to add optional start + duration numbers to nodes so we can import non-sampling data sources?

file-format-spec.ts
jlfwong: Speedscope currently imports from a wide variety of sampling profile formats (Chrome, Firefox, Instruments, Ruby stackprof, and anything that exports in the collapsed stack format). So this isn't just to export instruments files, but to export any sampling profile.

That said, your comment does suggest to me that there may be a better file format that would seamlessly support both sampled profiles & instrumented profiles. Let me mull it over and get back to you.

Thanks for the questions!

file-format-spec.ts
```diff
@@ -0,0 +1,87 @@
+// This file contains types which specify the speedscope file format.
+
+export namespace FileFormat {
+  export interface File {
+    version: string
+    exporter: 'https://www.speedscope.app'
```
jlfwong: I guess it's going to be weird to author JSON files that conform to this format but which are not actually exported by speedscope.

I'm considering having the file format be specified as a JSON schema (http://json-schema.org/). In that case, the `$schema` field could be used as a replacement for `exporter` to unambiguously identify a JSON ball as being a speedscope profile.

file-format-spec.ts
```diff
@@ -0,0 +1,87 @@
+// This file contains types which specify the speedscope file format.
```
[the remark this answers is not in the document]
uber-mahyar2: This is definitely a simpler format to create.  People who want to export to this format don't need to recreate the call tree anymore.  It definitely works with start/stop types of trace events.

How does a set of raw stack samples work with this format now?  I'm guessing your planning to export instruments files as start/stop spans vs a list of stack samples?  Or extend this later to export instruments-style stack samples?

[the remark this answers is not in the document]
jlfwong: > I'm guessing your planning to export instruments files as start/stop spans vs a list of stack samples?

Yeah, that was the idea. Anything currently imported into speedscope can be safely exported in this way right now, though it's perhaps a bit misleading since the actual input was sample-based, not event based.

An alternative approach would be to support both the original format I proposed and this new format. The door for doing that in the future would still be open, even if that isn't done in the initial version. That was the purpose of making `IProfile` a discriminated union of types.

In this current format, would this be useful for your immediate needs? The code in this PR both supports importing this format if you want to write tools that can output profiles to import into speedscope and supports exporting to this format if you wanted to use speedscope to e.g. import instruments profiles and export to an easier to parse format.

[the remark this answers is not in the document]
uber-mahyar2: Yeah this would work great.

Thank you so much for doing all of this!

PR #113, opened 2018-08-02
Make scripts/build-release.sh build a release build for CSP support

### LINKED ISSUE #108: Add a single file (or single JS file at least) build option
This would allow speedscope to be more easily integrated into environments where CSP is used

See: https://github.com/tmm1/stackprof/pull/100#issuecomment-409384898

jlfwong:
This should allow speedscope to be used in environments that have a `Content-Security-Policy` header which doesn't include `strict-dynamic`.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#strict-dynamic

Fixes #108


jlfwong:
Closing this for now since I think it's not actually needed.

PR #153, opened 2018-08-24
Change time formatting for minutes from 1.50min to 1:30

### LINKED ISSUE #90: Better labels for the time axis
First, it would really help if we could simplify how time > 1 minute is shown.  It's hard to parse 1.20min as 1:12 (have to constantly multiply the fraction by 60 to understand seconds). 

Second, only 2 decimals are shown, once you pass 1 minute, instead of 10ms as the finest grain now you only see 0.01 minutes (600ms) as the finest grain. To improve this it would be best if we could get a mm:ss.iii format (minutes:seconds.millisconds) instead. You can look at the timeline in Instruments to see a demo of what this would look like.

Thanks!

Alex-Diez:
Hi @jlfwong, I made changes to `TimeFormatter` to fix #90. Now it should render `1:30min` instead of `1.50min` 

However, I am not very proficient in Typescript and I did an ugly workaround. Maybe you can suggest something more elegant?

### DIFF
--- a/src/lib/value-formatters.test.ts
+++ b/src/lib/value-formatters.test.ts
@@ -7,7 +7,8 @@ describe('TimeFormatter', () => {
     expect(f.format(0.04)).toEqual('40.00µs')
     expect(f.format(3)).toEqual('3.00ms')
     expect(f.format(2070)).toEqual('2.07s')
-    expect(f.format(1203123)).toEqual('20.05min')
+    expect(f.format(150000)).toEqual('2:30')
+    expect(f.format(1203123)).toEqual('20:03')
   })
 
   test('input units seconds', () => {
@@ -15,7 +16,8 @@ describe('TimeFormatter', () => {
     expect(f.format(0.00004)).toEqual('40.00µs')
     expect(f.format(0.003)).toEqual('3.00ms')
     expect(f.format(2.07)).toEqual('2.07s')
-    expect(f.format(1203.123)).toEqual('20.05min')
+    expect(f.format(150)).toEqual('2:30')
+    expect(f.format(1203.123)).toEqual('20:03')
   })
 })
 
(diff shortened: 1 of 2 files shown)


src/lib/value-formatters.test.ts
```diff
@@ -7,15 +7,17 @@ describe('TimeFormatter', () => {
     expect(f.format(0.04)).toEqual('40.00µs')
     expect(f.format(3)).toEqual('3.00ms')
     expect(f.format(2070)).toEqual('2.07s')
-    expect(f.format(1203123)).toEqual('20.05min')
+    expect(f.format(150000)).toEqual('2:30min')
```
jlfwong: I think just "2:30" is okay rather than "2:30min"

src/lib/value-formatters.ts
```diff
@@ -25,7 +25,16 @@ export class TimeFormatter implements ValueFormatter {
   formatUnsigned(v: number) {
     const s = v * this.multiplier
 
-    if (s / 60 >= 1) return `${(s / 60).toFixed(2)}min`
+    if (s / 60 >= 1) {
```
jlfwong: I think you want something like this:

```
if (s / 60 >= 1) {
  const minutes = Math.floor(s / 60)
  const seconds = Math.floor(s - minutes * 60)
  return `${minutes}:${zeroPad(seconds, 2)}`.
}
```

Where `zeroPad` is a function that you can import from `utils.ts`: https://github.com/jlfwong/speedscope/blob/828beb7ccf329ad3d3f6821ea53e577621ec6432/src/lib/utils.ts#L84

You can use `const` instead of `let` here because the variable is never reassigned. 

You might already know this, but in case you don't, you can read about this here: https://wesbos.com/let-vs-const/.

Alex-Diez:
ping @jlfwong. If you are around a computer by this time 😄

jlfwong:
Thanks @Alex-Diez! 🎉

jlfwong:

PR #154, opened 2018-08-27
Allow optional CR before LF when probing collapsed stacks files

### LINKED ISSUE #152: Support Windows line endings (CRLF) in Brendan Gregg's collapsed stack format text files
It seems that the app chokes on text files with Windows line endings, with the generic "Unrecognized format! See documentation about supported formats." error popup. Unless it's deliberate, JWZ-style (https://www.jwz.org/xscreensaver/xscreensaver-windows.html), the support would spare Windows users the extra step of converting the line endings to LF.
[collapsed-stacks-with-crlf.txt](https://github.com/jlfwong/speedscope/files/2315968/collapsed-stacks-with-crlf.txt)
[collapsed-stacks-with-lf.txt](https://github.com/jlfwong/speedscope/files/2315969/collapsed-stacks-with-lf.txt)

januszn:
This fixes #152, in that it allows "collapsed stacks" files generated with
tools using Windows line endings to be imported into the tool verbatim.

### DIFF
--- a/sample/profiles/stackcollapse/simple-crlf.txt
+++ b/sample/profiles/stackcollapse/simple-crlf.txt
@@ -1,0 +1,5 @@
+a;b;c 1
+a;b;c 1
+a;b;d 4
+a;b;c 3
+a;b 5
--- a/src/import/__snapshots__/bg-flamegraph.test.ts.snap
+++ b/src/import/__snapshots__/bg-flamegraph.test.ts.snap
@@ -53,3 +53,57 @@ Object {
 exports[`importFromBGFlameGraph: indexToView 1`] = `0`;
 
 exports[`importFromBGFlameGraph: profileGroup.name 1`] = `"simple.txt"`;
+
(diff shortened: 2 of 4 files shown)


jlfwong:
Thanks for doing this!

I'll be happy to merge this as soon as a test covering this is added.

To add a test for this, you can add a test in this file: https://github.com/jlfwong/speedscope/blob/master/src/import/bg-flamegraph.test.ts

You can more or less copy the existing test in there, and add a version of https://github.com/jlfwong/speedscope/blob/master/sample/profiles/stackcollapse/simple.txt with all the `\n` replaced with `\r\n`.

Once you've added that test, if you run:

```
npm run jest -- -u
```

Then that should update the jest snapshot (https://jestjs.io/docs/en/snapshot-testing) including the expected output for that profile.

jlfwong:
Thanks so much for the reporting the issue, contributing the fix, and following through with rounds of feedback!

januszn:
No problem, I'm happy to help.

jlfwong:

PR #156, opened 2018-09-04
Match more Firefox-internal locations

jyc:
Looks like Firefox also generates locations with names like
`bound (self-hosted:951:0)`. We check for `self-hosted`, but not for
`self-hosted` with stuff after it following a colon. We should ignore
these too, otherwise we can end up with stuff on our stack that we don't
expect. This was causing Firefox profiles not to load because we
completed building the profile with a non-empty stack.

Attached is a profile that errors without this patch and successfully renders
with this patch.

[copy.json.zip](https://github.com/jlfwong/speedscope/files/2350583/copy.json.zip)


src/import/firefox.ts
```diff
@@ -180,7 +180,11 @@ export function importFromFirefox(firefoxProfile: FirefoxProfile): Profile {
 
         if (!match) return null
 
-        if (match[2].startsWith('resource:') || match[2] === 'self-hosted') {
+        if (
+          match[2].startsWith('resource:') ||
+          match[2] === 'self-hosted' ||
```
jlfwong: Isn't this second check redundant? I think this can just be:

```
if (match[2].startsWith('resource:') || match[2].startsWith('self-hosted:'))
```

Right?

jyc: I wasn't sure about why the `match[2] == self-hosted` check was added. I hypothesized that it worked in the past because older versions of Firefox or just Firefox in specific situations would generate locations with `self-hosted` and without a colon and other information following. I suppose `match[2].startsWith('self-hosted')` (no colon) would work too, but it feels a little weird.

jlfwong: Oh you're totally right. This is fine, sorry I misread

jlfwong:

PR #158, opened 2018-09-05
Pretty print JSON example

zoidbergwill:


jlfwong:
Thanks! This was output non-pretty-printed originally because that's how speedscope actually saves out this format. But I see this was linked from https://github.com/jlfwong/speedscope/wiki/Importing-from-custom-sources#speedscopes-file-format, which I agree means it is better served from being pretty-printed.

Issue #77, opened 2018-06-28
Format Support: Trace Event Format


uber-mahyar2:
Chrome has a simplified trace format which is a json list of trace dictionaries.  This is the main format we generate internally at our company, so it would be nice if speedscope could render them.

This test trace file loads in `chrome://tracing` while it gives an error in speedscope.

[test.json.zip](https://github.com/jlfwong/speedscope/files/2147191/test.json.zip)

uber-mahyar2:
Yeah it's the trace format in that doc.  

It looks very similar to this format that currently works: [Profile-20180628T170616.json.zip](https://github.com/jlfwong/speedscope/files/2147331/Profile-20180628T170616.json.zip)

I cut a much larger trace into that smaller one, since it worked in the chrome trace and it would be simpler to understand.  The bigger trace has a similar problem.  It comes from stack snapshots that we globbed together into spans for the chrome tracer, much like this item in the attached Profile.json:

```
  {
    "pid": 1996,
    "tid": 27395,
    "ts": 11937507063,
    "ph": "X",
    "cat": "toplevel",
    "name": "MessagePumpLibevent::OnLibeventNotification",
    "args": {
      "src_file": "../../mojo/edk/system/channel_posix.cc",
      "src_func": "StartOnIOThread"
    },
    "dur": 61,
    "tdur": 62,
    "tts": 178590489
  },
```

So the original data source it comes from are call trees.  When you give the chrome tracer the `"ph": "X"` type spans, chrome seems to expect they are call trees too, because the rendering goes pretty wonky if it isn't.  What chrome is rendering looks like it's a call tree to me: 

<img width="549" alt="screen shot 2018-06-28 at 5 34 42 pm" src="https://user-images.githubusercontent.com/24603300/42067167-b732f2e2-7af9-11e8-9bf8-0f0967ac892d.png">

But yes, it doesn't guarantee they are call trees.

jlfwong:
I’ll commit to implementing this feature myself if a total of $200 of donations is made to any combination of these specific charities:

- [American Civil Liberties Union](https://www.aclu.org/)
- [/dev/color](https://www.devcolor.org/welcome): /dev/color is a non-profit organization whose mission is to empower Black software engineers to help one another grow into industry leaders.
- [Girls Who Code](https://girlswhocode.com/about-us/): Girls Who Code was founded with a single mission: to close the gender gap in technology

[...]

PR #165, opened 2018-09-16
Add go tool pprof import support

### LINKED ISSUE #60: Format Support: Import from go pprof
There is Google protobuf-based file format used for profilers in Google tooling, and is more specifically used by `go tool pprof`.

https://github.com/google/pprof/tree/master/proto
https://blog.golang.org/profiling-go-programs

jlfwong:
This PR adds support for importing from Google's pprof format, which is a gzipped, protobuf encoded file format (that's incredibly well documented!) The [pprof http library](https://golang.org/pkg/net/http/pprof/) also offers an output of the trace file format, which continues to not be supported in speedscope to date (See #77). This will allow importing of profiles generated by the standard library go profiler for analysis of profiles containing heap allocation information, CPU profile information, and a few other things like coroutine creation information.

In order to add support for that a number of dependent bits of functionality were added, which should each provide an easier path for future binary input sources

- A protobuf decoding library was included ([protobufjs](https://www.npmjs.com/package/protobufjs)) which includes both a protobuf parser generator based on a .proto file & TypeScript definition generation from the resulting generated JavaScript file
- More generic binary file import. Before this PR, all supported sources were plaintext, with the exception of Instruments 10 support, which takes a totally different codepath. Now binary file import should work when files are dropped, opened via file browsing, or opened via invocation of the speedscope CLI.
- Transparent gzip decoding of imported files (this means that if you were to gzip compress another JSON file, then importing it should still work fine)

Fixes #60.

--

This is a [donation motivated](https://github.com/jlfwong/speedscope/issues/60#issuecomment-419660710) PR motivated by donations by @davecheney & @jmoiron to [/dev/color](https://www.devcolor.org/welcome) 🎉


jlfwong:
So far all this PR contains is a sample go program w/ some code to extract a profile from it, and a checked in version of that profile. Next I'll be working on extracting the profile data by add a dependency on some protocol buffer parsing library & trying to interpret it.

jlfwong:
There were some intermediary steps I forgot about but here's what I've done since the last update:

1. Refactored speedscope to be able to import from binary formats that are dropped on the window or opened, and to automatically detect files that are gzip compressed and decompress them.
2. Integrated [protobufjs](https://www.npmjs.com/package/protobufjs) as a tool to decode protocol buffers
3. Used `protobufjs`'s `pbjs` command to generate a JavaScript decoder for the [`profile.proto`](https://github.com/google/pprof/blob/master/proto/profile.proto) message type, and also `pbts` to generate TypeScript bindings for that file
4. Set up speedscope to log the decoded profile to the console.

Here's the result of dropping the `sample.prof` file into speedscope: 

image

Next up is going to be actually interpreting the results in order to start generating flamecharts/flamegraphs from the profile data!

PR #172, opened 2018-10-05
add base64 support to HashParams

ItsJimi:
Hello,

For pm2.io we want to be able to open one of our profiling from the app. For that, speedscope need to handle hash parameters in `#profileURL`. To not break compatibility we just added another hash paramater: `#profileBase64URL` like that you can give an url with query parameters without it being parsed by chrome.

Thanks

### DIFF
--- a/src/lib/hash-params.test.ts
+++ b/src/lib/hash-params.test.ts
@@ -24,4 +24,7 @@ test('getHashParams', () => {
   })
   expect(getHashParams('#abc=bcd')).toEqual({})
   expect(getHashParams('garbage')).toEqual({})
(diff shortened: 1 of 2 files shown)


jlfwong:
Hi @ItsJimi!

I don't think this change should be necessary because of this line:

https://github.com/jlfwong/speedscope/blob/d78d20e0057a17fe8cbf1a98e7e9292c8047a3eb/src/lib/hash-params.ts#L16

If you want to include hash params in the URL, then you just need to URL encode them. For instance:

```
encodeURIComponent("https://www.example.com/path#val=something")
> "https%3A%2F%2Fwww.example.com%2Fpath%23val%3Dsomething"
```

Also, did you mean query parameter (e.g. `?field=value`) rather than hash param? (e.g. `#hashparam`)? Hash parameters don't get received by the server at all, so I don't think it makes sense for speedscope to send them in the XHR.

Please let me know if I've misunderstood the problem.

Issue #176, opened 2018-10-08
chrome profile: reported time isn't the same as dev tools


vmarchaud:
I was working on support an older format of cpu profile when i noticed some difference between the values reported by speedscope and by devtools : 

![image](https://user-images.githubusercontent.com/2820968/46615803-70aff200-cb19-11e8-8cb8-cc729da78c06.png)

Both are ordered by self time but we see a lot of difference in the list. Also we don't report the idle time in speedscope, i believe we should ?

I'm willing to put some time to inspect it, just opening an issue to keep you up to date ;)

jlfwong:
It's possible this is a bug in the import, or a bug elsewhere in speedscope, though it isn't a specific goal to have this information map 1: w/ what devtools reports.

A specific example of something speedscope does that will result in a misalignment is to consider certain function calls to be part of whatever the previous stack was. As an example, the file format places `(garbage collector)` as a top-level frame in the profiler, but speedscope places it on top of whatever stack immediately precedes it. We do a similar thing for the opaque `(program)` calls.

See e.g. #85.

I don't confidently know offhand what would be causing this misalignment though, so it may be worth investigating and understanding it. It's possible however, that we'll discover it's an intentional different interpretation of the data.

vmarchaud:
Well the main problem for me is that if the devtools show it in this specific way, i believe its meant to be understand like this.
However we can see considerable difference between devtools and speedscope (for exemple the `idiff` total time, i will for sure inspect what's going on to be able to explain it

vmarchaud:
For merging some frames, i need some digging to better understand how its working. But if the thread is blocked when the GC is running, i believe speedscope should show it as a top level frame, and same for the `(program)` (which is [apparently ](https://stackoverflow.com/questions/3847954/what-is-program-in-chrome-debugger-s-profiler)v8 itself)

PR #177, opened 2018-10-09
Import v8 cpu profile (old format)

vmarchaud:
As said on #170, i added the support for the old format used by https://github.com/hyj1991/v8-profiler-node8 (which is currently used for pm2.io).

~~I also re-introduced the `(idle)` frame, and removed the merge between the `gc`/ `program` frame and the previous stack.~~

~~As discussed in #176, i believe speedscope need to show what's happening in the main thread at any given time, both of the `gc` and `program` frame represent blocking operation, so they should be shown at the top level so user can understand that their code isn't running while they are.~~

~~Note that `program` is actually native code, either V8 or any C++ binding :
Source : https://trac.webkit.org/changeset/138004/webkit~~


jlfwong:
> I also re-introduced the (idle) frame, and removed the merge between the gc/ program frame and the previous stack.

> As discussed in #176, i believe speedscope need to show what's happening in the main thread at any given time, both of the gc and program frame represent blocking operation, so they should be shown at the top level so user can understand that their code isn't running while they are.

I'm afraid I don't think I agree on either of these fronts. We can certainly discuss, but at a minimum I'd like these to be part of a different PR, since they affect more than what's described in this PR description, and I'd like to avoid interleaving comments about that change with comments about the correctness of the old format import.

vmarchaud:
Sure i'll do another PR tomorrow

jlfwong:
Thanks for splitting out #178! I'm pretty busy for the next few days, so as a heads up, it might take a while for me to get to review this & do a release

vmarchaud:
@jlfwong no problem we deployed a fork with all the fix/modification in our end anyway

src/import/index.ts
```diff
@@ -122,11 +124,14 @@ async function _importProfileGroup(dataSource: ProfileDataSource): Promise<Profi
       console.log('Importing as Firefox profile')
       return toGroup(importFromFirefox(parsed))
     } else if (isChromeTimeline(parsed)) {
-      console.log('Importing as Chrome CPU Profile')
+      console.log('Importing as Chrome Timeline')
```
jlfwong: Whoops! Nice catch and thanks for fixing this!

src/import/chrome.ts
```diff
@@ -264,3 +265,15 @@ export function importFromChromeCPUProfile(chromeProfile: CPUProfile): Profile {
   profile.setValueFormatter(new TimeFormatter('microseconds'))
   return profile.build()
 }
+
+/**
```
jlfwong: Thanks for writing this comment! I think this would more helpfully live at the top of `v8cpuFormatter.ts`.

I think this could also use a bit of rewording. Something like this:

```
* This importer handles an old format used by the C++ API of V8. This format is still used by v8-profiler-node8.
* There are two differences between the two formats:
*  - Nodes are a tree in the old format and a flat array in the new format
*  - Weights are timestamps in the old format and deltas in the new format.
*
* For more information, see https://github.com/hyj1991/v8-profiler-node8
```

Your English is still much, much better than my French 😄

Merci beaucoup pour tout votre contributions! 🇨🇦

src/import/v8cpuFormatter.ts
```diff
@@ -0,0 +1,61 @@
...
+    let lastTimestamp = index === 0 ? startTime * 1000000 : timestamps[index - 1]
+    deltas.push(timestamp - lastTimestamp)
+    return deltas
+  }, [])
+}
+
+/**
+ * Convert the tree based format to the array of nodes like actually
+ */
+export function chromeTree2nodes(content: OldCPUProfile): CPUProfile {
```
jlfwong: I generally haven't been using "2" to denote "to" in the code in this repository. I think I'd prefer "chromeTreeToNodes" here.

src/import/v8cpuFormatter.ts
```diff
@@ -0,0 +1,61 @@
...
+function timestampsToDelta(timestamps: number[], startTime: number): number[] {
+  return timestamps.reduce((deltas: number[], timestamp: number, index: number) => {
+    let lastTimestamp = index === 0 ? startTime * 1000000 : timestamps[index - 1]
+    deltas.push(timestamp - lastTimestamp)
+    return deltas
+  }, [])
+}
+
+/**
+ * Convert the tree based format to the array of nodes like actually
```
jlfwong: I'm not exactly sure what you were trying to express here, but perhaps something like this is what you intended?

```
/**
 * Convert the old tree-based format to the new flat-array based format
 */
```

src/import/v8cpuFormatter.ts
```diff
@@ -0,0 +1,61 @@
+import {CPUProfile, CPUProfileNode} from './chrome'
+
+export interface OldCPUProfileNode {
```
jlfwong: I don't think this needs exporting, does it?
```suggestion
interface OldCPUProfileNode {
```

src/import/v8cpuFormatter.ts
```diff
@@ -0,0 +1,61 @@
...
+    hitCount: node.hitCount,
+    children: node.children.map(child => child.id),
+  })
+  node.children.forEach(child => {
+    return treeToArray(child, nodes)
+  })
+  return nodes
+}
+
+function timestampsToDelta(timestamps: number[], startTime: number): number[] {
```
jlfwong: I think this would be better named as `timestampsToDeltas` (with "deltas" being plural)

src/import/v8cpuFormatter.ts
```diff
@@ -0,0 +1,61 @@
...
+  return timestamps.reduce((deltas: number[], timestamp: number, index: number) => {
```
jlfwong: It doesn't look to me like you need to use `reduce` here? Could this be more concisely expressed like so?

```
return timestamps.map((timestamp, index) => {
  const lastTimestamp = index === 0 ? startTime * 1000000 : timestamps[index - 1]
  return timestamp - lastTimestamp
})
```

src/import/v8cpuFormatter.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export interface OldCPUProfile {
+  startTime: number
+  endTime: number
+  head: OldCPUProfileNode
+  samples: number[]
+  timestamps: number[]
+}
+
+function treeToArray(node: OldCPUProfileNode, nodes: CPUProfileNode[]) {
```
jlfwong: This is a very minor thing, but I find functions that both mutate their arguments *and* return a new value confusing. I think this could be avoided like so:

```
function treeToArray(root: OldCPUProfileNode): CPUProfileNode[] {
  const nodes: CPUProfileNode[] = []
  function visit(node: CPUProfileNode) {
    nodes.push({
      id: node.id,
      callFrame: {
        columnNumber: 0,
        functionName: node.functionName,
        lineNumber: node.lineNumber,
        scriptId: node.scriptId,
        url: node.url,
      },
      hitCount: node.hitCount,
      children: node.children.map(child => child.id),
    })
    nodes.children.forEach(visit)
  }
  visit(root)
  return nodes
}
```

src/import/v8cpuFormatter.ts
```diff
@@ -0,0 +1,61 @@
...
+    deltas.push(timestamp - lastTimestamp)
+    return deltas
+  }, [])
+}
+
+/**
+ * Convert the tree based format to the array of nodes like actually
+ */
+export function chromeTree2nodes(content: OldCPUProfile): CPUProfile {
+  // Care that both startTime and endTime are now in microsecond
```
jlfwong: I think the more natural way of saying this would be:

```suggestion
  // Note that both startTime and endTime are now in microseconds
```

How would you say this in French? Attention?

src/import/chrome.ts
```diff
@@ -22,15 +23,15 @@ interface PositionTickInfo {
   ticks: number
 }
 
-interface CPUProfileCallFrame {
+export interface CPUProfileCallFrame {
```
jlfwong: I don't think this needs to be exported
```suggestion
interface CPUProfileCallFrame {
```

src/import/v8cpuFormatter.ts
vmarchaud: Yes, somethings among the lines of `Attention, les deux variables startTime et endTime sont en microseconds`.

vmarchaud:
Sorry for the one month delay, we were busy with our production :)
i believe i addressed everything you requested

jlfwong:
@vmarchaud Thanks for updating the PR! All the code LGTM, just need to get CI to pass.

Looks like the only thing causing CI failure is this:

```
/home/travis/build/jlfwong/speedscope/src/import/v8cpuFormatter.ts
  4:2  error  Delete `·`  prettier/prettier
  5:2  error  Delete `·`  prettier/prettier
  6:1  error  Delete `·`  prettier/prettier
  7:2  error  Delete `·`  prettier/prettier
  8:2  error  Delete `·`  prettier/prettier
  9:1  error  Delete `·`  prettier/prettier
```

I think you should be able to resolve this by running `npm run prettier` on this branch and checking in the resulting changes. If it worked, then `npm run lint` should exit cleanly.

PR #178, opened 2018-10-10
chrome: do not merge gc/program frame + show idle frame #176

vmarchaud:
As discused in https://github.com/jlfwong/speedscope/pull/177, i'm making a new PR for this change (sorry for not doing it directly, i was in rush to finish before leaving work)
Some arguments for the why : 

> As discussed in #176, i believe speedscope need to show what's happening in the main thread at any given time, both of the `gc` and `program` frame represent blocking operation, so they should be shown at the top level so user can understand that their code isn't running while they are.
Note that `program` is actually native code, either V8 or any C++ binding :
Source : https://trac.webkit.org/changeset/138004/webkit


vmarchaud:
Bumping this one too, i would like to understand why you believe it's better to merge some frames together (maybe i missed something ?)

jlfwong:
Sorry it took me a while to get to this!

I'm not sure how to produce an example for `(program)`, but here's an example showing that it's at least the right thing to do for `(garbage collection)`.

I made this simple program to force garbage collection to occur regularly:

```
<script>
function run() {
    const map = new Map();

    function addRandom() {
        map.set(Math.random(), new Uint8Array(100));
    }

    function removeRandom() {
        const keys = map.keys();
        const k = Array.from(map.keys())[Math.floor(Math.random() * map.size)]
    }

    for (let i = 0; i < 1000; i++) {
        addRandom()
    }

    for (let i = 0; i < 10000; i++) {
        if (Math.random() < 0.5) {
            addRandom()
        } else {
            removeRandom()
        }
    }
}

run()
</script>
```

When I profile this in Chrome, the profile looks like this in Chrome developer tools:

image

If I import into speedscope **without** the patch in this PR, I get this:

image

If I import into speedscope **with** the patch in this PR, I get this:

image

So Chrome itself does the same thing where it stacks the `(garbage collector)` frames on top of the current JavaScript stack.

I think there are similar ways of causing `(program)` stacks to appear in the middle of a JavaScript call stack, but I can't remember how.

vmarchaud:
I believe both are different things, you are using the `Performance` tab for Chrome which isn't only a CPU Profiler.
This tool implement a CPU Profiler and the tracing infrastructure of V8 (cf https://v8.dev/docs/trace).


The output of the CPU Profiler (which is what i'm trying to fix here), is always a file ending with `.cpuprofile`. The file that is generated by the Performance tab is always ending with `.json` (because it's an array of trace events). I believe you used a `.json` file in your first screenshot so that's why i believe it's not the same things.

Also i think it's shown like this in the Performance tab because the GC is run in parallel of the main thread (which is what V8 is doing most of the time nowadays), sadly there are no documentation about the devtools apart from their code itself.

In conclusion i think we could keep the non merging behavior for the CPU Profile and keep it for the trace events so it looks like the same as in V8.
What do you think ?

vmarchaud:
Could you please share with me the two profiles you made please ?

jlfwong:
@vmarchaud I think I'm confident that the behavior for `(garbage collector)` is correct in speedscope, and I haven't figured out a way to reproduce a `(program)` block to evaluate whether it's misleading or helpful. Given that, I'm going to close this PR for now. If you have new insights on the matter, please feel free to comment and we can discuss further.

PR #181, opened 2018-10-16
Onydra/hash param view mode

onydra:
Hi @jlfwong, thanks for this wonderful tool. This PR simply adds the ability to set the default viewMode from hash params so that one can directly link the most relevant view.


jlfwong:
I'm going to close this out for inactivity. Feel free to open if you'd like to see this landed once the requested changes have been addressed.

PR #184, opened 2018-10-27
Make the wasd keymappings work on azerty keyboards

vrischmann:
Instead of using `key`, use `code` which according to [this](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) should work consistently for different layouts.

I tested the modification on a french AZERTY keyboard and it works fine.


jlfwong:
Sweet, thanks!

jlfwong:
This is now merged, published to npm, and deployed to https://www.speedscope.app/!

vrischmann:
Nice, thank you !

PR #187, opened 2018-11-08
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
Node.js 6, 8 and 10 are current LTS branches. 11 is stable (`node`). 9 already reached its EOL. See https://github.com/nodejs/Release/blob/master/README.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
 language: node_js
 node_js:
-  - '9'
+  - 'node'
+  - '10'
+  - '8'
+  - '6'
+install: npm install


jlfwong:
Hi @DanielRuf!

This is fine, though these tests are actually exercising code that is ultimately only ever run in-browser, so compatibility across node version isn't really important.

The only bit of code that is actually run through node aside from the build system is https://github.com/jlfwong/speedscope/blob/master/bin/cli.js, which isn't exercised by these tests anyway.

If this is going to lead to longer CI feedback, I'd prefer to just use only latest LTS version (which sounds like it's 11).

DanielRuf:
Latest LTS is 10, odd releases like 11 are stable / short living.

Well but the dependencies need Node.js and can break.

DanielRuf:
And we can also use the cache of Travis CI to save some time.

jlfwong:
> Latest LTS is 10, odd releases like 11 are stable / short living.

Whoops, yes. That's definitely what you said originally, and I just misread.

> Well but the dependencies need Node.js and can break.

I'm not sure what this means. Which dependencies are you referring to? The only thing listed in `dependencies` is "opn", which isn't exercised on CI at all right now. Within the things listed within `devDependencies`, some of those are build/test tooling, and some of them only really need to work within browsers.

Did you mean that the build/test tooling like `eslint`, `jest`, and `parcel-bundler` could break?

As long as this woks with latest LTS and w/ stable, I'm not sure supporting old versions of node is important, since this only affects developers of speedscope itself.

Did you want to work on speedscope but encountered problems because of incompatibilites with the node version you were running? I guess I'm still not understanding what problem you're trying to solve with this PR.

DanielRuf:
Well, on 11 it is currently breaking and we test with the latest version locally using node-env.

https://github.com/jlfwong/speedscope/pull/187#issuecomment-436897882

6 and 8 can be probably removed, but 10 and latestshould be tested.

jlfwong:
Interesting! Having both 10 and 11 makes sense to me.

I'm not sure what's causing the test failures for 11 in https://travis-ci.org/jlfwong/speedscope/jobs/452253772. Is that something you were planning on investigating, or were you hoping I would look into that?

DanielRuf:
I'll try to find the cause / check it as far as I can and let you know what I find out.

jlfwong:
I'm going to close this out for inactivity. If someone would like to take this over and cares about this particular issue enough to get this to pass through travis-ci, please feel free to open a new PR.

DanielRuf:
I could check that. Can you reopen it?

jlfwong:
The "Reopen and comment" button says "The repository that submitted this pull request has been deleted", so it looks like I can't?

DanielRuf:
Ah ok. I forgot that I had cleaned up a bit and removed many forks. I will create a new PR and test on my Travis CI instance before opening it on your side.

PR #190, opened 2018-11-13
[DVSPIMP-2786] Dockerization.

rafaelbogaveev:
This is a link to recorded video https://drive.google.com/file/d/192hNlDCUOO4BmHiOw21fO1snGIQ09qxS/view?usp=sharing

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,34 @@
+# Dockerfile
+ Dockefile is created on top of `ubuntu` image.
+ # Requirements
+ - The project should be cloned from https://github.com/trilogy-group/speedscope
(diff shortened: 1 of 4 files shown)


jlfwong:
Hi! I, uh, don't really know what this is? It seems like a dockerized development environment for speedscope?

I'm also not sure if this was targeting the wrong branch? Was this just accidentally targeted to upstream instead of your fork? I see https://github.com/trilogy-group/speedscope/pull/1 is still open?

It looks like there's probably no action required here on my part -- is that right?

shubhdevf:
Hi Jamie,
Sorry about that. We hit the wrong fork.

PR #202, opened 2019-01-27
Support importing partial JSON files

loganek:
Partial files are allowed in many specs, e.g. Trace Event Format,
so the viewer should be able to load partial files as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
     "ts-jest": "22.4.6",
     "typescript": "2.8.1",
     "typescript-eslint-parser": "17.0.1",
-    "uglify-es": "3.2.2"
+    "uglify-es": "3.2.2",
+    "partial-json-parser": "1.0.3"
   },
   "jest": {
     "transform": {
--- a/sample/profiles/trace-event/simple-object-partial.json
+++ b/sample/profiles/trace-event/simple-object-partial.json
@@ -1,0 +1,7 @@
+{
+  "traceEvents": [
+    {"pid": 0, "tid": 0, "ph": "B", "name": "alpha", "ts": 0},
+    {"pid": 0, "tid": 0, "ph": "B", "name": "beta", "ts": 1},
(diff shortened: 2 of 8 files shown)


loganek:
Hey @jlfwong , could you please review that change? Thanks!

jlfwong:
@loganek Hi! My home internet has been down for the last week, but I'll try to get to this soon!

I'm generally wary of adding third party dependencies, but I took a look at `partial-json-parser`, and it looks like it has no dependencies and is enough code that I'd rather not just do it myself, so I'm happy to accept this.

I'll try to take a closer look later this week, merge if I don't find anything concerning, then do a release that picks this up.

Out of curiousity, were you motivated to do this because you have Chrome trace profiles that are not terminated? What was generating them?

loganek:
@jlfwong So sorry, I didn't want to rush you at all! Just thought that you missed the PR, so wanted to ping you, but please take your time, it's not super urgent for me.

Yeah, I wasn't sure what's your opinion about third party dependencies, but since it wasn't very time consuming, I decided to give a try.

Yes, I did this PR exactly because my files are not terminated; those files are generated by hawktracer-converter (https://github.com/loganek/hawktracer-converter/) which is a client app of my profiler (https://github.com/amzn/hawktracer). I guess I could work-around this in the converter to close the file properly, but since the format specification allow to have non-terminated files, I thought it'd be more correct to fix your viewer instead.

Btw. great tool, it feels much faster than the one embedded to chrome, and flamegraphs is a nice feature too. Good job, and thanks for doing that :)

jlfwong:
This is great as is :) Thanks for taking the time to carefully craft this PR!

jlfwong:
@loganek Ah... sorry I spoke too soon. I had to revert the commit because `partial-json-parser` doesn't guarantee termination: https://github.com/rgovind92/partial-json-parser/issues/3

PR #203, opened 2019-02-01
add default instruments selected run number

Archerlly:
this's will lack `com.apple.xray.owner.template` in instruments archive data where run instruments with command line.
like:
1. run`instruments -t Template.tracetemplate -D demo.trace -l 10000 -w  test.app`
2. drag `demo.trace` into `https://www.speedscope.app`
3. alert `Unrecognized format! See documentation about supported formats`


src/import/instruments.ts
```diff
@@ -361,7 +361,11 @@ async function readFormTemplate(tree: TraceDirectoryTree): Promise<FormTemplateD
   const archive = readInstrumentsKeyedArchive(await readAsArrayBuffer(formTemplate))
 
   const version = archive['com.apple.xray.owner.template.version']
-  const selectedRunNumber = archive['com.apple.xray.owner.template'].get('_selectedRunNumber')
+  let selectedRunNumber = 1
+  if ('com.apple.xray.owner.template' in Object.keys(archive)) {
```
jlfwong: I don't think this is quite right. The `in` test is for objects, not arrays, and `Object.keys` returns an array, so I think this condition will always be false.

image


Instead, I think you want:

```suggestion
  if ('com.apple.xray.owner.template' in archive) {
```

src/import/instruments.ts
```diff
@@ -361,7 +361,11 @@ async function readFormTemplate(tree: TraceDirectoryTree): Promise<FormTemplateD
...
+    selectedRunNumber = archive['com.apple.xray.owner.template'].get('_selectedRunNumber')
+    console.log('get selectedRunNumber', selectedRunNumber)
```
jlfwong: I'm guessing this was useful for debugging, but can you please remove this before we merge?

Archerlly:
> Hi @Archerlly,
> 
> Thanks for the submission!
> 
> Do you have a small `.trace` file that you can include as a test to ensure that this behavior doesn't regress in the future?
> 
> I should be able to create this myself, though I'm not sure where to find a `.tracetemplate` file to create this.

src/import/instruments.ts
Archerlly: - `.tracetemplate` file create by Instruments `File` -> `Save As Template`
- `console.log('get selectedRunNumber', selectedRunNumber)` is debug code and have removed
- add a testing with small `.trace` file that without selected run number

Thanks for your suggestion

Archerlly: - `.tracetemplate` file create by Instruments `File` -> `Save As Template`
- `console.log('get selectedRunNumber', selectedRunNumber)` is debug code and have removed
- add a testing with small `.trace` file that without selected run number

Thanks for your suggestion

Archerlly:
- remove `* console.log('get selectedRunNumber', selectedRunNumber)`
- remove `Object.keys` for `in` func
- add a testing with small `.trace` file that without selected run number

Thanks for your suggestion

jlfwong:
@Archerlly Your fix has now been published to npm and deployed to https://speedscope.app/. Thanks for the fix!

PR #210, opened 2019-02-18
Make Sandwich view resizable

### LINKED ISSUE #209: Resize frames in Sandwich view
Thanks for bringing this amazing product out!

Currently, the width of frames for (1) list of functions and (2) flamegraph of a selected function have fixed size, but it would be great if they can be resized by, for example, dragging the border of each frame.

okapies:
I try to implement #209 but I'm very new to TypeScript and don't see how to set up an environment and build the sources to check if it works. :cry: Is there a document like `BUILD.md`?

Fixes #209.

### DIFF
--- a/src/views/sandwich-view.tsx
+++ b/src/views/sandwich-view.tsx
@@ -1,10 +1,10 @@
 import {Frame} from '../lib/profile'
 import {StyleSheet, css} from 'aphrodite'
 import {ProfileTableViewContainer} from './profile-table-view'
-import {h} from 'preact'
+import {h, Component} from 'preact'
 import {commonStyle, Sizes, Colors, FontSize} from './style'
 import {actions} from '../store/actions'
-import {createContainer, Dispatch, StatelessComponent} from '../lib/typed-redux'
+import {createContainer, Dispatch} from '../lib/typed-redux'
 import {ApplicationState} from '../store'
 import {InvertedCallerFlamegraphView} from './inverted-caller-flamegraph-view'
 import {CalleeFlamegraphView} from './callee-flamegraph-view'
@@ -18,7 +18,53 @@ interface SandwichViewProps {
   glCanvas: HTMLCanvasElement
 }
 
-class SandwichView extends StatelessComponent<SandwichViewProps> {
+interface SandwichViewState {
+  isDragging: boolean
+  resizerPos: number | null
+  initialPos: number
+  delta: number
+}
+
+class SandwichView extends Component<SandwichViewProps, SandwichViewState> {
+  constructor(props: SandwichViewProps) {
+    super(props)
+    this.state = {
+      isDragging: false,
+      resizerPos: null,
+      initialPos: 0,
+      delta: 0,
+    }
+  }
+
+  startResize = (e: MouseEvent) => {
+    this.setState({
+      isDragging: true,
+      initialPos: e.clientX,
+      delta: 0,
+    })
+  }
+
+  resizeViews = (e: MouseEvent) => {
+    if (this.state.isDragging) {
+      let delta = e.clientX - this.state.initialPos
+      this.setState({delta: delta})
+    }
+  }
+
+  stopResize = (e: MouseEvent) => {
+    if (this.state.isDragging) {
+      let resizerPos = this.state.resizerPos
+      if (!resizerPos) {
+        resizerPos = this.state.initialPos
+      }
+      this.setState({
+        isDragging: false,
+        resizerPos: resizerPos + this.state.delta,
+        delta: 0,
+      })
+    }
+  }
+
   private setSelectedFrame = (selectedFrame: Frame | null) => {
     this.props.setSelectedFrame(selectedFrame)
   }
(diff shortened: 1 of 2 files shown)


jlfwong:
Sweet, thanks for starting to look into this! There is indeed a document: https://github.com/jlfwong/speedscope/blob/master/CONTRIBUTING.md

okapies:
Thanks! I missed it. :sweat_smile:

okapies:
I fixed to work resizing feature. It basically works, but it seems the moving amount is half of the expected... Any idea?

okapies:
Finally it seems it works! Please feel free to fix it to fit your taste.

jlfwong:
Sweet! I'll try to take a look at this sometime in the next week or so

src/views/sandwich-view.tsx
```diff
@@ -18,7 +18,53 @@ interface SandwichViewProps {
   glCanvas: HTMLCanvasElement
 }
 
-class SandwichView extends StatelessComponent<SandwichViewProps> {
+interface SandwichViewState {
+  isDragging: boolean
+  resizerPos: number | null
+  initialPos: number
+  delta: number
+}
```
jlfwong: Component state disappears when components unmount, which in the case of speedscope, will happen when you e.g. switch from sandwich view to another view. I think it would be helpful to persist resize state across view switches so you don't need to keep resizing it when you switch between views.

To deal with application-level state, speedscope uses redux (https://redux.js.org/). We store all of the application state in a single state object that lives longer than any of the components we unmount in speedscope's application lifecycle.

This means that rather than having the resize position stored in the `state` property of the `SandwichView`, it needs to be stored in the `SandwichViewState` object specified here: https://github.com/jlfwong/speedscope/blob/30ca6291ca93557e2bb77a2bafc5bac580627102/src/store/sandwich-view-state.ts#L10-L12

If you look at the diff I put in the top-level review feedback comment, you can see how I did that.

This part of the architecture of speedscope is admittedly not very well documented :)

src/views/sandwich-view.tsx
```diff
@@ -18,7 +18,53 @@ interface SandwichViewProps {
...
+
+class SandwichView extends Component<SandwichViewProps, SandwichViewState> {
+  constructor(props: SandwichViewProps) {
+    super(props)
+    this.state = {
+      isDragging: false,
+      resizerPos: null,
+      initialPos: 0,
+      delta: 0,
```
jlfwong: Rather than having multiple nullable properties & booleans expressing a single idea, I like to group together this information into a single concept where none of the properties are nullable. You can see that in the diff I attached to the top-level review comment:

```diff
+interface ActiveDrag {
+  lastMouseX: number
+}
+
 interface SandwichViewState {
-  isDragging: boolean
-  resizerPos: number | null
-  initialPos: number
-  delta: number
+  activeDrag: ActiveDrag | null
 }
```

This is an alternative structure to e.g. doing:

```
interface SandwichViewState {
  lastMouseX: number | null
  isDragging: boolean
}
```

and I think it's preferable, because then you don't have to consider the case of `isDragging` being `true` while `lastMouseX` is `null`.

src/views/sandwich-view.tsx
```diff
@@ -18,7 +18,53 @@ interface SandwichViewProps {
...
+    if (this.state.isDragging) {
+      let resizerPos = this.state.resizerPos
+      if (!resizerPos) {
+        resizerPos = this.state.initialPos
+      }
+      this.setState({
+        isDragging: false,
+        resizerPos: resizerPos + this.state.delta,
+        delta: 0,
+      })
```
jlfwong: As far as I can tell from this code, the resizing only happens on mouse up. I think it's a much better user experience to have continuous feedback during dragging.

src/views/sandwich-view.tsx
```diff
@@ -30,19 +76,50 @@ class SandwichView extends StatelessComponent<SandwichViewProps> {
...
+    let resizer: JSX.Element | null = null
     let flamegraphViews: JSX.Element | null = null
 
     if (selectedFrame) {
+      functionsView = (
+        <div
+          className={css(style.tableView)}
+          style={{
+            width: `calc(100% - ${this.state.resizerPos ? this.state.resizerPos + 'px' : '50%'})`,
+          }}
```
jlfwong: Because `functionsView`, `resizer`, and `flamegraphViews` are all children of a container with `display: flex`, it shouldn't be necessary to specify all of their sizes changing during the resize. Instead, it should be sufficient only specify one and have the others "flex" to fill the remaining space.

In the code I attached to the top-level comment, I chose to only make `flamegraphViews` resize, and let the `functionsView` "flex" to fill the remaining space.

src/views/sandwich-view.tsx
```diff
@@ -64,22 +141,42 @@ class SandwichView extends StatelessComponent<SandwichViewProps> {
...
         <div className={css(style.tableView)}>
           <ProfileTableViewContainer activeProfileState={this.props.activeProfileState} />
         </div>
+      )
+    }
+
+    container = (
+      <div
+        className={css(commonStyle.hbox, commonStyle.fillY, style.flexDisplay)}
+        onMouseUp={e => this.stopResize(e)}
```
jlfwong: I think this is covered by the binding of `mouseup` on `window`, so I think this is unnecessary.

src/views/sandwich-view.tsx
```diff
@@ -30,19 +76,50 @@ class SandwichView extends StatelessComponent<SandwichViewProps> {
...
+          style={{left: this.state.delta}}
+          onMouseDown={e => this.startResize(e)}
+        />
+      )
       flamegraphViews = (
-        <div className={css(commonStyle.fillY, style.callersAndCallees, commonStyle.vbox)}>
+        <div
+          className={css(commonStyle.fillY, style.callersAndCallees, commonStyle.vbox)}
+          style={{
+            width: `calc(100% - ${this.state.resizerPos ? this.state.resizerPos + 'px' : '50%'})`,
```
jlfwong: Resizing the view isn't sufficient to update the flamechart view, resulting in it not re-rendering until mouse over. This creates visual artifacts that are avoidable by (somehow) forcing the flamechart view to re-render.

Currently the `FlamechartPanZoomView` listens for window resize events to re-render:

https://github.com/jlfwong/speedscope/blob/30ca6291ca93557e2bb77a2bafc5bac580627102/src/views/flamechart-pan-zoom-view.tsx#L699

https://github.com/jlfwong/speedscope/blob/30ca6291ca93557e2bb77a2bafc5bac580627102/src/views/flamechart-pan-zoom-view.tsx#L404-L407

I haven't tried this, but one way to fix this might be to fake a window resize event, like this:

```
window.dispatchEvent(new UIEvent('resize'))
```

I didn't fix this in the code I wrote and attached to the top-level PR

okapies:
I'm very sorry for the late response. Thanks for your comments! I learned a lot from them. But I'm a bit busy to do my own work (analyzing and improving the performance of our software using speedscope) and don't have enough time to fix them now. I'd like to come back again sometime but wonder if you could complete it if you feel this feature useful.

jlfwong:
I'm going to close this out for inactivity. If anyone would like to take this up in the future, please feel free to open another PR with the requested changes addressed.

PR #215, opened 2019-03-07
Fix parse NSString/NSMutableString always fail in keyedArchive file

Archerlly:
Like parese `com.apple.xray.owner.template.description` value always none

### DIFF
--- a/src/import/instruments.ts
+++ b/src/import/instruments.ts
@@ -740,7 +740,7 @@ function paternMatchObjectiveC(
       // Replace NSString with a string
       case 'NSString':
       case 'NSMutableString':
-        return decodeUTF8(value['NS.bytes'])
+        return decodeUTF8(value['NS.bytes']) || value['NS.string']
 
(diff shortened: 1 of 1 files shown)


Archerlly:
> Nice! Do you have an example where import fails without this?

1. parse test file in `sample/profiles/Instruments/10.0/simple-time-profile.trace`
2. the func `readInstrumentsKeyedArchive` parse archive with key `com.apple.xray.owner.template.description` always none

jlfwong:
Sorry for never following up with this! I ended up needing this change to fix some build issues, so thank you for filing it!

A nearly identical change was included as part of another PR: https://github.com/jlfwong/speedscope/pull/253/files#diff-639b339f6c7404035aca21a584e9a76fR743-R746

I'll credit you in the CHANGELOG.

jlfwong:
This should be fixed in 1.5.3 on npm, and on https://www.speedscope.app/. Thanks again for your contribution!

PR #233, opened 2019-10-04
Support newer Emscripten .symbols with hex escapes

jyc:
Apparently Emscripten now generates `.symbols` files where names are not mangled using Clang's mangling scheme, but rather hex-escaped! So 'a\20b' means 'a b'. Currently we can't import these symbol maps into Speedscope because a regex rejects them, and they look weird because we don't unescape.


src/lib/emscripten.ts
```diff
@@ -4,7 +4,7 @@ type EmscriptenSymbolMap = Map<string, string>
 //
 // NOTE: This will fail to ignore escaped backslahes (e.g. `\\20`).
 function unescapeHex(input: string): string {
-  return input.replace(/\\([a-fA-F0-9]{2})/, (_match, group) => {
+  return input.replace(/\\([a-fA-F0-9]{2})/g, (_match, group) => {
```
jlfwong: Eugh, JavaScript regexes are the worst regexes. Why is the regex stateful :<

jyc:
Will merge and try to deploy now!

src/lib/emscripten.ts
jyc: 😢

Issue #191, opened 2018-11-15
Do not ellipsize file name


Yajo:
When hovering over calls, the file & method is ellipsized to the left, which is useless in my case (and I imagine also for others):
![captura de pantalla de 2018-11-15 11-05-05](https://user-images.githubusercontent.com/973709/48549061-a4272e80-e8c6-11e8-8406-4b62e9d91269.png)

It shouldn't be ellipsized, or at least it should be to the right, because I think it's more interesting the file name and function name than the system path where it is found.

jlfwong:
Hi @Yajo!

I've heard this feedback from a few people now, and I think at least one version of it should be pretty easy to fix.

The relevant code to fix this is here: https://github.com/jlfwong/speedscope/blob/ad49dacb29f6b9dab22171af9da92e9b44abf909/src/views/hovertip.tsx#L11.

I'd accept a PR to make the tooltip extend to the edge of the container with a margin

miso11:
Hi, is enough to only increase TOOLTIP_WIDTH_MAX to 900?

jlfwong:
@miso11 Sure, I'd accept a PR to do that, assuming it never overflows the window (which I don't think it should)

jlfwong:
I'll close this for now, though I'd also be happy to take a more complete PR that does ellipsization in the middle of the string rather than the end

PR #239, opened 2019-10-23
Do not ellipsize file name

miso11:
Issue #191

It shouldn't be ellipsized, or at least it should be to the right, because I think it's more interesting the file name and function name than the system path where it is found.

Make max width bigger

Before: 
before

After:

after


jlfwong:
Hi @miso11! Just as good practice for understanding the effect of the PR, can you please include a screenshot of the before/after of this change in the PR description?

Thanks for the change!

miso11:
Pictures was added to PR description

jlfwong:
Sorry for the huge delay! This change is now published to npm & to https://www.speedscope.app/. Thanks again for your contribution!

Issue #260, opened 2020-04-19
Automate dependency updates


JustinBeckwith:
Greetings!  I noticed that a lot of dependencies are out of date, and it's preventing the codebase from building on Node.js 13.x.  One of the things we use to help keep deps up to date is renovate:
https://renovatebot.com/

I like it because it submits PRs for each dependency, making it easier to update one thing at a time.  Might help reduce the maintenance burden!

bjorn3:
Just curious, but what is the difference between renovatebot and the github owned dependabot?

JustinBeckwith:
At a glance, Dependabot looks great!  I only suggested renovate because it's the tool I'm most familiar with.

jlfwong:
Hi!

I don't have really strong feelings here, but I generally like to avoid taking on dependencies until I personally repeatedly face problems I'd like to automate away. The majority of the pain from dependency upgrades isn't finding versions to bump to, it's fixing all the problems that are caused by incompatibilities, so I'm not sure what I get out of this.

The fact that it doesn't build on Node.js 13.x is concerning to me, but can you please file that as a separate issue?

Issue #16, opened 2018-04-14
Allow speedscope to be used as a library instead of a standalone application


alangpierce:
I'm making a chrome devtools extension to surface the results of server-side profiling, and I tried hacking speedscope in as a React component, but I ran into some issues. From my perspective, it would be great if I could install speedscope as a library and render it as a React component with the full text of the profile as a prop.

Here's a partial list of issues that would need to be addressed for this to happen:
* You need to actually build it as a library and publish it on npm. 😄My hacky approach was to just `npm link` the TypeScript into my node_modules and compile it with webpack, but that ran into various issues. If you want to avoid the React dependency (or peer dependency), you could ship a `@speedscope/core` and separate `@speedscope/react` package, or something like that.
* Speedscope uses Preact, which isn't drop-in compatible with React. Probably the best way to ship a React-friendly speedscope component is to just write a React wrapper that mounts a Preact render tree inside, rather than trying to make the whole thing a React render tree.
* I ran into CSP issues since [regl](https://github.com/regl-project/regl) uses eval. I was able to enable eval for my extension to work around it, but I'd prefer not to. This seems like it might be unreasonable to work around, but it would be a good thing to mention in docs.
* Formatting looked wrong because speedscope includes reset.css, which it turns out is necessary. The aphrodite styles just worked, though, which was nice.
* The profile rendered correctly and various interactions worked, but hover and pinch zoom didn't do anything, and drag-to-pan caused the main flamechart view to just disappear. This might just be due to React/Preact differences, not sure.

So at least for now, it looks like an iframe (or external window) is the nicest way to use speedscope, and from what I can see that works fine. But it would be cool if it could be more integrated into a larger tool.

rmccue:
I just want to add a strong +1 to this.

We have a custom dashboard built in React which allows developers to take requests to servers and dive into the performance/etc of it. We have our own flamegraph tool currently (built upon https://github.com/bvaughn/react-flame-graph) which has the advantage of being deeply integrated into the dashboard, but which lacks all of the power and ease of use of Speedscope.

I'd love to be able to integrate it as a React component and just drop it in; it looks like potentially we could replicate what the `views/` directory's code does to integrate it, although I'm not sure if that's a use case you're looking to support. I'd really like to avoid iframeing it in if at all possible (we'd want to integrate the toolbar UI into our app natively).

[...]

PR #261, opened 2020-04-19
Unpin dev dependencies and address build failures

JustinBeckwith:
This change adds the `^` range specifier for many dev dependencies. The only exception is TypeScript, which likes to make breaking changes in semver minor releases.  I noticed that this projects doesn't build on Node.js 13.x, largely due to out of date dependencies.  This felt like a good first step to start upgrading dev dependencies one by one.

Note: the vast majority of changes here are either in `package-lock.json` or a result of running `npm run prettier`.


jlfwong:
Hi @JustinBeckwith!

> This change adds the ^ range specifier for many dev dependencies. The only exception is TypeScript, which likes to make breaking changes in semver minor releases. I noticed that this projects doesn't build on Node.js 13.x, largely due to out of date dependencies. This felt like a good first step to start upgrading dev dependencies one by one.

Can you explain a bit of the motivation for this? I strongly prefer to pin dependencies, especially dev dependencies for projects that aren't usable as libraries because it alleviates an entire class of reproducibility problems. It's really common for things that weren't intended to be breaking changes to be accidentally breaking changes, resulting in different installations happening on different people's machines (or even multiple machines where I'm running on both).

Bumping versions because of node incompatibility sounds fine to me, but I'd rather bump from pinned versions to pinned versions.

I understand the need for libraries which things directly depend upon to avoid version pinning to prevent massive duplication of similarly versioned packages and to pick up security fixes automatically, but neither of those are relevant for speedscope.

JustinBeckwith:
Greetings!  And thanks for the really thoughtful reply.  Just to be clear, whatever approach you want to take here is great, and there are no wrong answers.  You made a lovely thing, and I'd like to use it with a later version of node.js.  To share a little on why I general recommend folks go down the path of `^`....

The saving grace here is `package-lock.json`.   That file ensures the tree of transitive dependencies is locked for users that clone the repository.  Without doing something the explicitly would update the versions in that lock file, everyone _should_ be working from the same locked tree of top level and transitive dependencies.  This has no effect on the packages in the shipped module.  The benefit is you get to accept bug fixes, security fixes, and features of libraries as they come out.   I suspect you would still be just as safe due to the lock file.  

Either way to decide to go - I do really like having a bot bump my dependencies for me (see #260).  If you decide to keep dev dependencies pinned down, tools like [renovate](https://github.com/renovatebot/renovate) can automatically submit PRs to bump them for you as updates come up.  With pinned dependencies, the volume of those updates could be pretty high, so you can even set it up to automerge PRs in particular situations (dev deps only, CI passes, patch only, etc).

Just to make sure I don't lose the forest through the trees - the entire goal was to start inching towards supporting later versions of node.js.  I tried to clone this repo, and `npm install` fails on the latest build of node.js 13.  The reason turned out to be an old pinned build of `jest`, which had an old version of `fs-events`, which was struggling to do some sort of native compilation.  The problem went away if I upgraded to the latest version of `jest`.  

P

JustinBeckwith:
Also - Apologies if it seemed like I jumped straight to solutions without filing the original root issue.  Thanks :)
https://github.com/jlfwong/speedscope/issues/262

jlfwong:
Thanks for filing #262! I'll see if I can try to fix that later today (possibly by merging something really similar to this PR). The specific context around jest is helpful.

Re: "The saving grace here is package-lock.json"

What happens when you manually bump one of the versions of something in `package.json`? Will the existence of `package-lock.json` ensure that *only* the version of something whose explicit version string was changed is updated, or may it also look for newer versions of packages with `^` or similar modifiers marked?

In any case, I find reasoning about the effects of changing `package.json` to be much easier with locked versions. `package-lock.json` is definitely helpful nonetheless to pin transitive dependencies (I've been in situations before where a coworker had a broken build despite all direct dependencies matching because of a version mismatch on a transitive dependency).

Thanks for the patient explanations, and thanks for cloning speedscope and trying to build it locally and letting me know it's busted on Node 13! I'm curious to know what you're trying to modify if you can share.

JustinBeckwith:
On bumping something in `package.json` - after you bump it, and run `npm install`, it will go through the whole package resolution process.  It will attempt to update the top level package, and only _if needed_ bump transitive dependencies if allowed within range.  There's a big complicated range resolution algorithm that has to find the highest common denominator across given ranges.  

On what I'm doing - of course!  A couple of years ago I wrote a thingy that lets you track the require time of a given npm module:
- https://npmtrace.appspot.com/
- https://github.com/JustinBeckwith/npmtrace/

It used the now-defunct trace viewer from chrome dev tools:
- https://github.com/catapult-project/catapult

A few weeks ago, after a chrome update, it just, stopped working.  I needed to get in and try updating to the latest version, but then I found this and figured why not 🤷   I like that this project lives on GitHub, is written in TypeScript, and looks to be pretty well maintained.  

So what I'd really like to do is pick up the trace viewer as a component, and have an API to load a trace directly given the JSON.  I have the backend service that does all the tracing and storage and API and such, I really just need to visualize it within my app.  
 

I have no idea what specifically I was going to try and change; I got distracted when I couldn't run `npm install` the first time 😆

jlfwong:
> A couple of years ago I wrote a thingy that lets you track the require time of a given npm module:

Cool!

> So what I'd really like to do is pick up the trace viewer as a component, and have an API to load a trace directly given the JSON. I have the backend service that does all the tracing and storage and API and such, I really just need to visualize it within my app.

Gotcha!

There are two mechanisms that might be interesting to you to do this. One that exists and one that I'm open to. The easiest is if you have a file that can be served from a URL, then you can specify the profile URL like so: https://github.com/jlfwong/speedscope#importing-via-url

The one that I'm open to is to use speedscope as a library. The discussion for that is taking place in #16.

Issue #262, opened 2020-04-19
`npm install` fails on node.js 13


JustinBeckwith:
Greetings!  I tried to clone the repository, and run `npm install` with the latest release of node.js 13.x.  The install fails:

```
$ node -v
v13.13.0

$ npm install
<.... lots of errors ....>
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/assert.h:93:25: note: 
      expanded from macro 'assert'
    (__builtin_expect(!(e), 0) ? __assert_rtn(__func__, __FILE__, __LINE__, #e) : (void)0)
                        ^
17 errors generated.
make: *** [deasync.target.mk:131: Release/obj.target/deasync/src/deasync.o] Error 1
make: Leaving directory '/Users/beckwith/Code/speedscope/node_modules/deasync/build'
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/Users/beckwith/.nvm/versions/node/v13.13.0/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:194:23)
gyp ERR! stack     at ChildProcess.emit (events.js:315:20)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:275:12)
gyp ERR! System Darwin 19.4.0
gyp ERR! command "/Users/beckwith/.nvm/versions/node/v13.13.0/bin/node" "/Users/beckwith/.nvm/versions/node/v13.13.0/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /Users/beckwith/Code/speedscope/node_modules/deasync
gyp ERR! node -v v13.13.0
gyp ERR! node-gyp -v v5.1.0
gyp ERR! not ok 
Build failed
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules/chokidar/node_modules/fsevents):
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 install: `node install`
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: Exit status 1

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! deasync@0.1.13 install: `node ./build.js`
npm ERR! Exit status 1
npm ERR! 
[...]

PR #254, opened 2020-01-16
Test on Node 10, 12 and latest

DanielRuf:
It seems the builds for 12 and latest fail because of different snapshots.

PR #263, opened 2020-04-20
Fix the build for node 13.x, make travis test 10, 12, 13, stable

jlfwong:
@JustinBeckwith pointed out in #262 that `npm install` was broken in node 13.x, and @DanielRuf pointed in #254 that test fail for node 11+ because of a change to stability of sorting.

This PR seeks to address both of those.

The installation issue was fixed by just regenerating `package-lock.json` without needing to bump any of the direct dependency versions. The test failure issue requires manual intervention.

To fix the sort stability issue, I updated the tests to use the stable sort values (these were all the correct values, though some of the test values were incorrect).

To make the suite still pass for node 10, I added a hack where I override `Array.prototype.sort` with a stable implementation that's *only* used in tests (See comments in code for a justification for why)

## Test Plan

Before this PR: `npm install` on node 13.x fails & `npm run jest` results in test failures
After this PR: `npm install` on node 13.x passes & `npm run jest` passes for node 10, 12, and 13.


.travis.yml
```diff
@@ -1,3 +1,5 @@
 language: node_js
 node_js:
-  - '9'
+  - '10'
+  - '12'
```
JustinBeckwith, line 4: Maybe worth explicitly adding 13 as well?

jlfwong: Sure

DanielRuf: `node` or `latest` may be better than `13` (fail early).

jlfwong: I added both

jlfwong:
This should be good to go, but Travis is down for scheduled maintenance, so I'm going to wait for it to come back up to make sure this gets through CI.

PR #264, opened 2020-04-20
Bump parcel and audit fix

JustinBeckwith:
This is the outcome of running `npm audit fix`.  Largely bumping transitive dependencies in the tree.


jlfwong:
Thanks!

PR #275, opened 2020-05-24
Try converting from preact to react

jlfwong:
I as was curious how much bigger the output bundle would be with React vs Preact, and also to see if there were other perf differences.

Using `scripts/prepare-test-installation.sh`, the main bundle is 208KB with preact, and 324K with react + react-dom. So it definitely is a non-trivial size difference.

Post-gzip, the difference is 54K vs 90K.


jlfwong:
I'm not going to do this for now since there isn't too much upside at the moment.

PR #277, opened 2020-05-24
WIP: Start trying to port to recoil

jlfwong:
I might abandon this, but I was curious what it would take to port speedscope to use recoil instead.

I'm currently stuck on thinking about how to do per-profile, per-speedscope-tab state that doesn't leak memory or create a ton of boilerplate. I'm not really clear on what granularity of atoms to use.


jlfwong:
Superseded by #341.

PR #281, opened 2020-05-26
Fix bug with bad caching of action creators

### LINKED ISSUE #280: Profile switching within a single multi-profile broken
I think this regressed in #267.

I haven't deployed this to the live site yet, so it isn't broken there.

jlfwong:
Profile switching was subtly broken because action creators weren't being correctly re-bound due to a missing dependency in a `useCallback` call.

I also tried to reduce boilerplate in this PR by adding additional exhaustive deps protection via eslint for `useSelector`, `useAppSelector`, and `useActionCreator`. The removes the need for using `useCallback` or each of those.

Fixes #280

### DIFF
--- a/src/views/flamechart-view-container.tsx
+++ b/src/views/flamechart-view-container.tsx
@@ -3,7 +3,6 @@ import {FlamechartID, FlamechartViewState} from '../store/flamechart-view-state'
 import {CanvasContext} from '../gl/canvas-context'
 import {Flamechart} from '../lib/flamechart'
 import {FlamechartRenderer, FlamechartRendererOptions} from '../gl/flamechart-renderer'
-import {ActionCreator} from '../lib/typed-redux'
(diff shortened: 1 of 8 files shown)


src/views/flamechart-view-container.tsx
```diff
@@ -28,51 +26,34 @@ interface FlamechartSetters {
...
-  function useActionCreatorWithIndex<T, U>(
-    actionCreator: ActionCreator<WithFlamechartContext<U>>,
-    map: (t: T) => U,
-  ): (t: T) => void {
-    const callback = useCallback(
-      (t: T) => {
-        const args = Object.assign({}, map(t), {id})
-        return actionCreator({profileIndex, args})
-      },
-      [actionCreator, map],
```
jlfwong, line 48: The actual bug was that `profileIndex` and `id` were missing from this array. I'm not sure why this wasn't caught by the exhaustive deps check.

Issue #139, opened 2018-08-16
Support remapping profiles with sourcemaps


jlfwong:
This would work similarly to the emscripten remapping. See, e.g. #93

PR #306, opened 2020-08-06
WIP: load sourcemaps

cricklet:
This is WIP! Let me know what you think of the overall approach and any guideline on testing. This PR is pretty much gluing `speedscope` to Mozilla's `source-map` project. Interestingly, they use a wasm binary to do the decoding quickly.

Another possible approach could be doing the source-map decoding manually, eg like: https://gist.github.com/bengourley/c3c62e41c9b579ecc1d51e9d9eb8b9d2.

I tested this with a tiny test project here: https://gist.github.com/cricklet/0deaaa7dd63657adb6818f0a52362651

## Before:

Screen Shot 2020-08-05 at 8 22 14 PM

## After:

Screen Shot 2020-08-05 at 8 22 07 PM

Fixes: https://github.com/jlfwong/speedscope/issues/139

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
   },
   "dependencies": {
     "opn": "5.3.0",
-    "react": "^16.13.1"
+    "react": "^16.13.1",
+    "source-map": "^0.7.3"
   }
 }
--- a/src/lib/js-source-map.ts
+++ b/src/lib/js-source-map.ts
@@ -1,0 +1,45 @@
+import * as sourcemap from 'source-map'
+import {MappingItem} from 'source-map'
+
+type JavascriptSymbolMap = Map<number, Map<number, string>>
+
+// Looks like the d.ts description doens't properly define `initialize`
+// @ts-ignore
+sourcemap.SourceMapConsumer.initialize({
+  'lib/mappings.wasm': 'https://unpkg.com/source-map@0.7.3/lib/mappings.wasm',
+})
+
+export async function importJavascriptSymbolMap(
+  contentsString: string,
+): Promise<JavascriptSymbolMap | null> {
+  try {
+    const contents = JSON.parse(contentsString)
+
+    if (contents.names == null) {
+      return null
+    }
+
+    if (contents.mappings == null) {
+      return null
+    }
+
+    const result = new Map<number, Map<number, string>>()
+
+    await sourcemap.SourceMapConsumer.with(contents, null, consumer => {
+      consumer.eachMapping(function (m: MappingItem) {
+        const line = m.generatedLine - 1
+        const col = m.generatedColumn + 1
+        if (!result.has(line)) {
+          result.set(line, new Map<number, string>())
+        }
+        result.get(line)?.set(col, m.name)
+      })
+    })
+
+    return result
+  } catch (e) {
+    return null
+  }
+
+  return null
+}
--- a/src/views/application.tsx
+++ b/src/views/application.tsx
@@ -13,6 +13,7 @@ import {LeftHeavyFlamechartView, ChronoFlamechartView} from './flamechart-view-c
 import {CanvasContext} from '../gl/canvas-context'
 import {Graphics} from '../gl/graphics'
 import {Toolbar} from './toolbar'
+import {importJavascriptSymbolMap} from '../lib/js-source-map'
 
 const importModule = import('../import')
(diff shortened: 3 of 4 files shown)


cricklet:
Hmm should I commit the `package.lock`?

src/views/application.tsx
```diff
@@ -13,6 +13,7 @@ import {LeftHeavyFlamechartView, ChronoFlamechartView} from './flamechart-view-c
 import {CanvasContext} from '../gl/canvas-context'
 import {Graphics} from '../gl/graphics'
 import {Toolbar} from './toolbar'
+import {importJavascriptSymbolMap} from '../lib/js-source-map'
```
jlfwong, line 16: I try to use code-splitting fairly aggressively in speedscope to keep the JS needed to boot the application as small as possible. To get this to be a part of a different chunk (which can be lazy-loaded), you can use the pattern below with `const importModule = import('../import')`. So in this case:

```
const jsSourceMapModule = import('../lib/js-source-map')
jsSourceMapModule.then(() => {})
```

And then put that below the `importModule` bits below to load it with a lower priority than the `import.ts` module.

This gives you a promise which resolves as the module. So you should be able to something like the following inline inside of the `loadFromFile` function:

`const {importJavascriptSymbolMap} = await jsSourceMapModule`

package.json
```diff
@@ -80,6 +80,7 @@
   },
   "dependencies": {
     "opn": "5.3.0",
-    "react": "^16.13.1"
+    "react": "^16.13.1",
+    "source-map": "^0.7.3"
```
jlfwong, line 84: Oh whoops, neither of these should be dependencies, they should only be `devDependencies` because it's needed to build speedscope but isn't needed to run speedscope as an npm module (the npm module only exposes a binary, not a JavaScript API).

I'm actually not sure why `react` is in here at all 😬 , but that's a mistake.

So as part of this PR, can you remove `react`, move `source-map` to `devDependencies`, and also commit the changes to `package-lock.json`?

src/lib/js-source-map.ts
```diff
@@ -0,0 +1,45 @@
+import * as sourcemap from 'source-map'
+import {MappingItem} from 'source-map'
+
+type JavascriptSymbolMap = Map<number, Map<number, string>>
```
jlfwong, line 4: Super nitty nit, but `JavaScriptSymbolMap` please, and similarly `importJavaScriptSymbolMap`.

src/lib/js-source-map.ts
```diff
@@ -0,0 +1,45 @@
...
+
+// Looks like the d.ts description doens't properly define `initialize`
+// @ts-ignore
+sourcemap.SourceMapConsumer.initialize({
+  'lib/mappings.wasm': 'https://unpkg.com/source-map@0.7.3/lib/mappings.wasm',
```
jlfwong, line 9: Woah this is pretty zany. I'm not sure what this is doing exactly, but it should be possible to pull this in as an import somehow? speedscope currently works offline (except that the fonts are different), so I'd like to keep it fully self-contained if possible (you can literally run speedscope from an extracted zip file with no internet connection).

src/views/application.tsx
```diff
@@ -223,11 +224,29 @@ export class Application extends StatelessComponent<ApplicationProps> {
...
+            profiles: [profile],
+          }
+        }
+
+        const jsMap = await importJavascriptSymbolMap(fileContents)
+        if (jsMap) {
+          const {profile, index} = this.props.activeProfileState
+          console.log('Importing as javascript symbol map')
+          profile.remapNamesForLineAndColumn(frame => {
+            console.log(frame.line, frame.col, jsMap.get(frame.line ?? -1)?.get(frame.col ?? -1))
```
jlfwong, line 244: I think this is part of this being WIP, but please remember to remove!

src/lib/js-source-map.ts
jlfwong: It seems like it should be possible to import the wasm module and have it be base64 encoded somehow. A tiny bit of information available here: https://github.com/mozilla/source-map/issues/408. It's possible that doing this will require mucking with the packaging config (I'm using https://parceljs.org/ at the moment, though have plans to switch to esbuild).

jlfwong:
Closing since #317 has now been landed!

Issue #168, opened 2018-09-27
Doesn't open browser in Windows


janpio:
```
C:\Projects\Fastlane\optimize_startup_time (master -> origin)
λ speedscope C:\Users\Jan\.cache\rbspy\records\rbspy-2018-09-27-Y7AYEI5ihd.speedscope.json
Creating temp file C:\Users\Jan\AppData\Local\Temp\speedscope-1538089620601-55152.js
Creating temp file C:\Users\Jan\AppData\Local\Temp\speedscope-1538089620601-55152.html
Opening file://C:\Users\Jan\AppData\Local\Temp\speedscope-1538089620601-55152.html in your default browser
```

Nothing happens.

Opening `file://C:\Users\Jan\AppData\Local\Temp\speedscope-1538089620601-55152.html` manually worked.

jlfwong:
Hi @janpio, thanks for reporting this! I've done verrry close to zero testing on windows, so this doesn't particularly surprise me :)

I'll try to find some time to investigate this on a Windows VM.

Out of curiousity, which shell are running `speedscope` in on Windows? Is this powershell or git bash or something else?

janpio:
Yeah, it didn't really surprise me as well, so no harm done ;) 
Easiest solution is probably to change the text on Windows that the user should open it manually.

I was using [Cmder](http://cmder.net/) right now, that usually behaves a bite more like git bash or a macOS or Linux shell than Powershell or cmd.exe. But should work everywhere of course 🔫

spillerrec:
Changing the line `cli.js:92`:
```JS
  await opn(urlToOpen, {wait: false})
```
to `wait: true` solves the issue for me both in cmd and Powershell.

jlfwong:
@spillerrec Oh sweet, thanks for figuring that out. I'd accept a PR to switch packages & patch the bug if you're interested!

spillerrec:
I have created a PR here: https://github.com/jlfwong/speedscope/pull/307
I haven't worked on Node projects before, so let me know if I managed to mess it up despite the small amount of changes.

jlfwong:
This should be fixed now on master thanks to @spillerrec, and will go out with the next release. No promises on when that's going to be, but hopefully sometime in the next week or two.

jlfwong:
This has now been released, so it should be fixed in `speedscope@1.10.0`.

PR #307, opened 2020-08-25
#168 Fix browser not opening on Windows when using the CLI

spillerrec:


jlfwong:
Thank you!! I'll hopefully have some time this weekend to test this out locally and then merge it.

bin/cli.js
```diff
@@ -89,7 +89,7 @@ async function main() {
 
   console.log('Opening', urlToOpen, 'in your default browser')
 
-  await opn(urlToOpen, {wait: false})
+  await open(urlToOpen, {wait: true})
```
jlfwong, line 92: It's really odd to me that `wait: true` is required to make this work on windows, but I'll just have to take your word for it.

On macOS anyway, `wait: false` is a nicer behaviour. So I think at least on macOS, I'd prefer if we retained `wait: false`. I'm not sure what the right thing do on linux is, but I'd like to conservatively assume that `wait: false` works too. So I think the best change here would be:

```suggestion
  // We'd like to avoid blocking the terminal on the browsing closing,
  // but for some reason this doesn't work at all on Windows if we
  // don't use wait: true.
  const wait = process.platform === "win32";
  await open(urlToOpen, {wait})
```

What do you think?

tmr232: I just checked on my machine (Windows 10) - it seems that just switching from the old `opn` to the new `open` is enough to fix the issue. No need to change the `wait` from what I can tell. Maybe that was only needed in the old version?

spillerrec: It doesn't seem to do the trick for me on both my Windows 10 machines. On the other hand, `await` does work if the browser is not already open and causes the script not to exit. I didn't notice that before since I always have Chrome open with a bunch of tabs and in this case speedscope exits without waiting for the browser to close.

I have not been able to test it properly though. `npm run serve` opens speedscope in a different way using a local server which have no issues, and `npm install -g` fails due to trying to run a bash script. So I have in fact been editing the global installation to test things. I might be missing the obvious since I don't develop in node, but I don't know how to properly start the local version in the same way the global installed version does. (In case this makes a difference.)

jlfwong: `npm run serve` is indeed totally different. To test the cli in dev, you'll need to run the following in the speedscope repository root:

```
# clone the repo, then...
npm install
npm run build  # This creates the necessary dist/release/index.html files
bin/cli.js
```

tmr232: Oh, my bad then. I didn't realize.

I get `npm ERR! missing script: build` when trying to run `npm run build`. I'll look into it tomorrow.

jlfwong: Ah whoops, that's my bad -- try `npm run prepack`.

tmr232: OK

Being a bash script, `npm run prepack` fails on Windows. I ran the commands manually.

So yeah, it only works with `wait: true`. Which is super weird - as it works perfectly well if I use `wait: false` when calling `open` from the node shell.

spillerrec:
I would have liked to look into what is going on in the `open` package, but I will have to admit that it is not going to happen. I believe the issue is with the `open` package but this workaround, while not being perfect, I think is better than it not working at all on Windows.

jlfwong:
Thank you!!

PR #319, opened 2020-10-11
Update link to Austin instructions

P403n1x87:
The Austin format conversion tools have been moved to the dedicated austin-python module. The README has been updated to point to the new instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ speedscope is designed to ingest profiles from a variety of different profilers 
 - Python
   - [Importing from py-spy](https://github.com/jlfwong/speedscope/wiki/Importing-from-py-spy-(python))
   - [pyspeedscope](https://github.com/windelbouwman/pyspeedscope)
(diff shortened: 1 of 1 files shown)


jlfwong:
Thanks!

jonashaag:
@P403n1x87 given that Speedscope supports the Austin format natively, why are we recommending users to go through the `austin2speedscope` script?

P403n1x87:
@jonashaag Whilst it's true that you can upload the output generated by Austin directly onto Speedscope for a quick visualisation, using the native format should be the "right" thing to do. For example, if you turn on full mode with the `-f` switch, `austin2speedscode` will generate all the profiles (time, memory allocations, memory deallocations) and separate them per process and thread.

**Aside**: Austin 3 adds a metadata header that breaks the currently deployed version of Speedscope, so simply uploading won't work, unless that header is manually stripped out first. This was "fixed" with #336.

PR #336, opened 2021-03-06
Allow collapsed stacks with invalid lines

P403n1x87:
Ingest files containing collapsed stacks and tolerate invalid lines, as FlameGraph does.

Some files might contain lines starting with a # to add comments to the collected samples. Speedscope should still attempt to parse these files as collapsed stacks and only keep the samples that it can find. Only fail if there are no samples reported.


jlfwong:
Hi @P403n1x87!

Before landing this, this needs tests for the cases it covers. You can see tests for similar traces here: https://github.com/jlfwong/speedscope/blob/main/src/import/bg-flamegraph.test.ts

P403n1x87:
> Before landing this, this needs tests for the cases it covers. You can see tests for similar traces here: https://github.com/jlfwong/speedscope/blob/main/src/import/bg-flamegraph.test.ts

I've copied and augmented the `simple.txt` case with some invalid lines. The generated snapshot looks fine. Please let me know if there's more that needs to be done. Cheers!

jlfwong:
Thanks!

PR #338, opened 2021-03-06
Allow paste in search

### LINKED ISSUE #337: Impossible to paste into search popup
When trying to paste something into the search popup, Speedscope tries to interpret the data in the clipboard as profile data. When the popup has the focus, the paste action should just paste the clipboard content into the search text field.

P403n1x87:
Fixes #337


jlfwong:
Thanks!

PR #353, opened 2021-06-15
[Proposal] Migrate to fast-check

dubzzz:
**What?**
This Pull Request is mostly a proposal to migrate from jsverify to fast-check.

**Why?**
jsverify's owner stopped the maintenance of the library more than a year ago. In addition, the last real commit to jsverify was on the 6th of September 2019.

**Benefits of `fast-check` over `jsverify`**
1. You don't need to define unmapping fuctions on your calls to `map` (`smap` in `jsverify`).
2. The arbitraries are able to find more bugs. For instance if you take a look into the updated tests you'll see that I had to calm down the generator of numbers because it was reporting issues if we used it fully.
3. There is a built-in way to handle commands. But I did not adapt the tests of LRUCache yet (they can definitely use the builder of commands provided by fast-check).
4. Maintained.


src/lib/lru-cache.test.ts
```diff
@@ -90,11 +90,13 @@ type Command =
     }
   | {type: 'removeLRU'}
 
-const arbitraryCommand: jsc.Arbitrary<Command> = jsc.record({
-  type: jsc.elements(['has', 'get', 'getSize', 'insert', 'getOrInsert', 'removeLRU']),
-  key: jsc.elements(['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h']),
-  value: jsc.nat,
-}) as any
+const arbitraryCommand = fc.record({
```
dubzzz, line 93: There is a built-in way to deal with commands in fast-check. I could probably migrate the test to it.

src/lib/math.test.ts
```diff
@@ -1,26 +1,35 @@
...
+let numericType = fc
+  .double({
+    next: true,
+    noDefaultInfinity: true,
+    noNaN: true,
+    min: -1 / numericTypeMinAccuracy,
+    max: 1 / numericTypeMinAccuracy,
   })
-})
+  .map(d => Math.round(d / numericTypeMinAccuracy) * numericTypeMinAccuracy)
```
dubzzz, line 15: Here the arbitrary `fc.double` was able to find potential issues when used with full resolution. Contrary to `jsc.number` it can generate any number from -Infinity to +Infinity including -0, Number.MIN_VALUE, Number.EPSILON and others (from small to large values).

src/lib/utils.test.ts
```diff
@@ -215,17 +221,19 @@ test('objectsHaveShallowEquality', () => {
...
+      for (let byte of byteArray) {
+        binaryString += String.fromCharCode(byte)
+      }
+      const b64string = btoa(binaryString)
+      const decoded = decodeBase64(b64string)
+
+      expect(Uint8Array.from(byteArray)).toEqual(decoded)
+
+      // If the above expect(...) assertion fails, we won't reach here.
+      return true
```
dubzzz, line 236: No need for return with fast-check. No return or return true is success. Throw or return false is failure.

dubzzz:
Cc @jlfwong, jsverify is clearly dead today. Any blocker to switch to fast-check? More maintained, more powerful, more used...

jlfwong:
Hey @dubzzz! If you want to resolve the conflicts on this and get it through CI, I'm happy to merge a PR.

I didn't take a careful look at this in the past because, despite jsverify being dead, that wasn't (and tbh, isn't) causing me any real pain, nor does this resolve any user facing issues.

All that said, I agree it makes more sense to use a library that's maintained in an ongoing manner (especially if the creator is willing to do the migration for you 😄)

sapnabehel67-jpg (changes requested):
coba dulu

dubzzz:
Oops I totally forgot about this one

PR #364, opened 2021-09-22
Add `README-zh_CN.md` file

fltenwall:
Add Chinese documents to support Chinese


jlfwong:
Hi @fltenwall! This is great -- thank you for contributing this!

I can't read Chinese, but the google translate for this seems very reasonable. When I update the README in the future, I'll tag you if you'd like to make corresponding changes to the Chinese.

Thanks!

fltenwall:
@jlfwong Sure, I'd love to.

PR #365, opened 2021-09-27
add file and line information

sokra:
* add file and line to tooltips
* add file and line to anonymous methods

## Before:

image

image

While you was able to get this info with clicking in "Time Order" and "Left Heavy" view, it was impossible to receive in the sandwich view.

## After:

image

image


jlfwong:
Hi @sokra! Thanks for all the work you've done towards building the modern JS ecosystem!

> While you was able to get this info with clicking in "Time Order" and "Left Heavy" view, it was impossible to receive in the sandwich view.

Ahh, this is a good call out. 

I think I'd rather see this approached slightly differently.

1. Always include file & line number in the hovertip
2. In the import codepath from Chrome which may have these `(anonymous)` frames, rename the symbol to also include the file & line number. The motivation for this is that speedscope imports from many formats, and `(anonymous)` is specific to Chrome profiles AFAIK.

What do you think?

sokra:
> What do you think?

Sounds good. I've updated the PR

shadeMe:
@jlfwong Any chance of merging this for the next release? Would be a welcome addition :)

Issue #373, opened 2021-11-01
Optimize the name display on the stack


YiniXu9506:
It seems not convenient to see the thread name if each stack showing full path of the thread name. 
![image](https://user-images.githubusercontent.com/34967660/139614737-f43b5654-9c9e-42fe-a210-853d97ea711d.png)

Expecting to show the last part of the thread name instead of full path, and users can also see the full path if they hover on the specific stack.
![image](https://user-images.githubusercontent.com/34967660/139616208-4b6f5a5c-d806-4300-8217-3813ffe070af.png)

YiniXu9506:
#374

PR #374, opened 2021-11-04
chore: make name display more concise on the stack

YiniXu9506:
This PR is related to issue #373 , which expecting to make name display more concise on the stack

### DIFF
--- a/src/lib/text-utils.ts
+++ b/src/lib/text-utils.ts
@@ -30,6 +30,9 @@ interface TrimmedTextResult {
 // Trim text, placing an ellipsis in the middle, with a slight bias towards
 // keeping text from the beginning rather than the end
 export function buildTrimmedText(text: string, length: number): TrimmedTextResult {
+  // make thread name more concise by spliting full path and showing the last part as thread name
+  const textFullPath = text.split("/")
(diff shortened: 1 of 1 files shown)


src/lib/text-utils.ts
```diff
@@ -30,6 +30,9 @@ interface TrimmedTextResult {
 // Trim text, placing an ellipsis in the middle, with a slight bias towards
 // keeping text from the beginning rather than the end
 export function buildTrimmedText(text: string, length: number): TrimmedTextResult {
+  // make thread name more concise by spliting full path and showing the last part as thread name
+  const textFullPath = text.split("/")
+  text = textFullPath[textFullPath.length - 1]
```
breeswish, line 35: Considering that speedscope accepts not only Golang style pprof data, but also other kind of profiling data from other languages, this might not be a general solution I think.

YiniXu9506: Yes, you'r right. I will go through all the supported languages and find a more general and better display way.

jlfwong:
Hi! Thanks for taking a look at this. This issue is quite old now, and I agree it doesn't generalize across profile format types. I'm going to close this for now.

PR #375, opened 2021-11-04
chore: ignore stack with negative sample

YiniXu9506:


Lauszus:
I have the same issue.

I worked around it by replacing all the negative weights with 0 before loading it:

```python
import argparse
import json

parser = argparse.ArgumentParser()
parser.add_argument("input")
parser.add_argument("output")
args = parser.parse_args()

input_: str = args.input
output: str = args.output

with open(input_) as f:
    data = json.load(f)

for i, p in enumerate(data["profiles"]):
    for j, w in enumerate(p["weights"]):
        if w < 0:
            data["profiles"][i]["weights"][j] = 0

with open(output, "w") as f:
    json.dump(data, f)

```

PR #383, opened 2021-12-15
fix: shouldn't merge (program) call frame

hyj1991:
Hi, thanks for this great work!

I am able to use speedscope to analyze `.cpuprofile`, but sometimes it behaves differently with devtools:

> [CPU-20200515T093903.cpuprofile.zip](https://github.com/jlfwong/speedscope/files/7719038/CPU-20200515T093903.cpuprofile.zip)

`CPU-20200515T093903.cpuprofile` in speedscope:

image

And in devtools the flamegraph is like:

image

---

> https://github.com/jlfwong/speedscope/blob/main/src/import/chrome.ts#L205-L207

Maybe we shouldn't to merge the `(program)` frame to keep the same behavior  with chrome devtools.

### DIFF
--- a/src/import/chrome.ts
+++ b/src/import/chrome.ts
@@ -203,7 +203,7 @@ function shouldIgnoreFunction(callFrame: CPUProfileCallFrame) {
 }
 
 function shouldPlaceOnTopOfPreviousStack(functionName: string) {
-  return functionName === '(garbage collector)' || functionName === '(program)'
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
This is something I'm running into frequently myself. Tried out the changes in the PR locally and they address the issue. With this PR the Chrome profiles look correct again.

marvinhagemeister (approved):

PR #385, opened 2022-03-27
Support importing profiles whose contents exceed V8s maximum string size

### LINKED ISSUE #340: Failed to load file(linux perf script) with size lager than 600MB
Failed to load file(linux perf script) with size lager than 600MB, thrim the size to around 500MB works fine.
Is this a bug or by design ?

jlfwong:
Browsers have a limit on how big you can make strings. In Chrome on a 64 bit machine, this is around 512MB, which explains why in #340, a 600MB file fails to load.

To work around this issue, we avoid making strings this large.

To do so, we need two core changes:
1. Instead of sending large strings as the import mechanism to different file format importers, we introduce a new `TextFileContent` interface which exposes methods to get the lines in the file or the JSON representation. In the case of line splitting, we assume that no single line exceeds the 512MB limit.
2. We introduce a dependency on https://github.com/evanw/uint8array-json-parser to allow us to parse JSON files contained in `Uint8Array` objects

To ensure that this code doesn't code rot without introducing 600MB test files or test file generation into the repository, we also re-run a small set of tests with a mocked maximum string size of 100 bytes. You can see that the chunked string representation code is getting executed via test coverage.

Fixes #340


package.json
```diff
@@ -53,6 +53,7 @@
     "ts-jest": "24.3.0",
     "typescript": "4.2.3",
     "typescript-json-schema": "0.42.0",
+    "uint8array-json-parser": "^0.0.1",
```
bjorn3, line 56: Why would adding a single dependency add 15000 lines to the package-lock.json file?

jlfwong: Huh. It's a good question.

My guess is something to do with the `lockfileVersion` change: https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion

Before this change, the `package-lock.json` was `lockfileVersion: 1`, and after the chang eit was `lockfileVersion: 2`. This has nothing to do with this change, it just so happens this is the first time I've installed something since the npm version change.

PR #388, opened 2022-04-26
add support for JS Self-Profiling data

nornagon:
This adds support for importing data in the format produced by the [JS Self-Profiling API](https://wicg.github.io/js-self-profiling/).

I'm not too familiar with the internals of this project, so I went the cheap route: there's a fairly straightforward mapping from the Self-Profiling data format to the Speedscope SAMPLED file format, so I've done that mapping and delegated to the Speedscope importer to do the rest of the work.

### DIFF
--- a/src/import/index.ts
+++ b/src/import/index.ts
@@ -22,6 +22,7 @@ import {decodeBase64} from '../lib/utils'
 import {importFromChromeHeapProfile} from './v8heapalloc'
 import {isTraceEventFormatted, importTraceEvents} from './trace-event'
 import {importFromCallgrind} from './callgrind'
+import {importFromJSSelfProfiling} from './js-self-profiling'
 
 export async function importProfileGroupFromText(
   fileName: string,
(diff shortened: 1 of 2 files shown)


jlfwong:
Hi @nornagon!

Thanks for adding this -- I didn't know this API existed, but it's so cool that it does!

The code as-is certainly isn't the best way of doing it. You probably want to use `StackListProfileBuilder`, like we do here: https://github.com/jlfwong/speedscope/blob/main/src/import/bg-flamegraph.ts

Before this can merge, this would also need tests. You can see example tests here: https://github.com/jlfwong/speedscope/blob/main/src/import/bg-flamegraph.test.ts

Given that this is also a new profile type, this would also ideally include the files necessary to generate this kind of profile locally, in case other features are added to the profile filetype. You can see example programs e.g. in https://github.com/jlfwong/speedscope/tree/main/sample/programs/javascript.

Which brings me to a question: I tried generating one myself locally to understand this, and ran into errors that say something like `[Violation] Document policy violation: js-profiling is not allowed in this document.` It seems like this is fixable via setting headers from the server. How have you been generating profiles for this?

PR #391, opened 2022-05-11
Support stackprof object mode

alexcoco:
First off, apologies for not opening and issue to discuss this ahead of time as recommended in the contributing guide -- I kind of worked on this on a whim this morning.

This PR attempts to support stackprof's object mode which tracks the number of allocated objects. This differs from the other modes (cpu and wall) by taking samples every time a Ruby object is allocated using Ruby's [`NEWOBJ` tracepoint](https://github.com/tmm1/stackprof/blob/df24b85953bb45d3abff58d9c82169a3003a60f1/ext/stackprof/stackprof.c#L198-L199).

When importing an object mode profile into speedscope today it still works but what you see is a profile using time units. The profile will only have samples for when an object was allocated which means even if time is reported, the profile is not really meaningful when looking at time.

To address this I've done three things when `mode` is `object`:
+ adjusted the total size of the `StackListProfileBuilder` to use the number of samples (since each sample is one allocation)
+ adjusted the weight of each sample to be `nSamples` (which I believe is always `1` but I'm not positive)
+ do not set the value formatter to a time formatter

Here's what it looks like before and after my changes (note the units and weight of samples):

wall (before) | object (before) | object (after)
-- | -- | --
Screen Shot 2022-05-11 at 4 51 31 PM | Screen Shot 2022-05-11 at 4 51 34 PM | Screen Shot 2022-05-11 at 4 51 42 PM

<details>
<summary>Test code</summary>

```ruby
require 'stackprof'
require 'json'

def do_test
  5.times do
    make_a_word
  end
end

def make_a_word
  ('a'..'z').to_a.shuffle.map(&:upcase).join
end

StackProf.start(mode: :object, interval: 1, raw: true)
do_test
StackProf.stop

File.write('tmp/object_profile.json', JSON.generate(StackProf.results))

StackProf.start(mode: :wall, interval: 1, raw: true)
do_test
StackProf.stop

File.write('tmp/wall_profile.json', JSON.generate(StackProf.results))
```
</details>

Finally, I am not super familiar with TypeScript and even less familiar with this codebase so I'm not sure what tests, if any, are needed for a change like this. I'm happy to add whatever else is needed. 🙇


jlfwong:
Hi @alexcoco! Thanks for adding this -- this seems totally reasonable to me.

Before this can land, this needs tests to cover the new codepath. You can see how the other test for stackprof works by looking here: https://github.com/jlfwong/speedscope/blob/main/src/import/stackprof.test.ts

That test file was generated, I believe, using the following test file: https://github.com/jlfwong/speedscope/blob/main/sample/programs/ruby/simple.rb

Please include the ruby file you used to profile, and a new profile. Both the program and the resulting profile should be as short as possible while still exercising the core functionality.

Let me know if you need any help!

Speedscope is badly in need of a new release. I'm planning on doing that sometime in the next month.

alexcoco (commented):
I added a sample program and profile along with a test case to run it. It generated a snapshot which seems to be committed as well so I've included that too.

Here are the before (speedscope.app) & after (local server) versions of this profile:

Before | After
-- | --
Screen Shot 2022-05-16 at 4 20 16 PM | Screen Shot 2022-05-16 at 4 20 20 PM

Issue #393, opened 2022-05-16
perf data idle time


prim:
speedscope provides a great time-based view.

But for perf data, when the program is idle, speedscope will think that the time is consumed by the next sampled call stack. This is wrong, the program should be idle at this time.

```text

$ cat test.c

#include <stdio.h>
#include <unistd.h>
#include <time.h>
#include <pthread.h>
#include <sys/time.h>
#include <sys/types.h>

void fselect() {
    fd_set rfds;
    struct timeval tv;
    int retval;

    FD_ZERO(&rfds);
    FD_SET(0, &rfds);
    tv.tv_sec = 2;
    tv.tv_usec = 0;
    retval = select(1, &rfds, NULL, NULL, &tv);

    if (retval == -1)
        perror("select()");
    else if (retval)
        printf("Data is available now.\n");
        /* FD_ISSET(0, &rfds) will be true. */
    else
        printf("No data within two seconds.\n");
}

void timed(int microseconds) {
    int sec, usec, diff;
    struct timeval tv;
    gettimeofday(&tv, NULL);
    // printf("begin seconds : %ld micro seconds : %ld\n", tv.tv_sec, tv.tv_usec);
    sec = tv.tv_sec;
    usec = tv.tv_usec;

    while (1) {
        gettimeofday(&tv, NULL);
        diff = (tv.tv_sec - sec) * 1000000 + tv.tv_usec - usec;
        if (diff > microseconds) {
            // printf("end seconds : %ld micro seconds : %ld\n", tv.tv_sec, tv.tv_usec);
            break;
        }
    }
}

void a() { timed(500000); }
void b() { timed(500000); }
void c() { timed(500000); }
void d() { timed(500000); }
void e() { timed(500000); }

void *fthread(void *args) {
    func();
    return NULL;
}

void func() {
    while (1) {
        a();
        b();
        fselect();
        c();
        d();
        e();
        printf("sleep begin\n");
        sleep(1.5);
        printf("sleep end\n");
    }
}

int main() {
    pthread_t thread;
    pthread_create(&thread, NULL, fthread, NULL);

    func();
    return 0;
}

$ gcc test.c -O0 -g -lpthread && ./a.out

$ perf record -F 99 -p 320355 -g -e cpu-clock 

```

The result now looks like this:

![image](https://user-images.githubusercontent.com/1023487/168552027-7f876f61-106a-49c8-a426-089d45d8b02a.png)

A more correct result should look like this:

[...]

PR #401, opened 2022-07-31
WIP: Try to get the pthreads idle example running under docker

jlfwong:
Working towards fixing #393

### DIFF
--- a/sample/programs/cpp/perf-script.sh
+++ b/sample/programs/cpp/perf-script.sh
@@ -14,6 +14,14 @@ if [[ "$1" == "forks" ]]; then
   exit 0
 fi
 
+if [[ "$1" == "idle" ]]; then
+  perf record -a -F 999 -g ./idle > perf.data
+  echo "Done recording"
+  ls -alh perf.data
+  perf script -i perf.data
+  exit 0
+fi
+
 perf record -F 999 -g ./simple-terminates > perf.data
 
 if [[ "$1" == "with-header" ]]; then
--- a/sample/programs/cpp/Dockerfile
+++ b/sample/programs/cpp/Dockerfile
@@ -1,12 +1,11 @@
 FROM ubuntu:bionic
 
 RUN apt-get update
-RUN apt-get install -y build-essential
-RUN apt-get install -y linux-tools-common linux-tools-generic
+RUN apt-get install -y build-essential linux-tools-common linux-tools-4.15.0-30-generic libpthread-stubs0-dev
 RUN ln -sf /usr/lib/linux-tools/4.15.0-30-generic/perf /usr/bin/perf
 
 WORKDIR /workdir
 
 ADD . /workdir
 
-RUN make clean simple-terminates forks
+RUN make clean simple simple-terminates forks idle
--- a/sample/programs/cpp/Makefile
+++ b/sample/programs/cpp/Makefile
@@ -17,9 +17,15 @@ forks: forks.cpp
 random: random.cpp
 	$(CXX) -Wall -g -o $@ $<
 
+idle: idle.c
+	$(CC) -Wall -g -pthread -o $@ $<
+
 linux-perf:
+	docker container prune -f
+	docker image rm speedscope-ubuntu || true
 	docker build -t speedscope-ubuntu .
-	docker run --privileged speedscope-ubuntu /bin/bash perf-script.sh forks > ../../profiles/linux-perf/forks.linux-perf.txt
+	docker run --privileged speedscope-ubuntu /bin/bash perf-script.sh idle > ../../profiles/linux-perf/idle.linux-perf.txt
+	# docker run --privileged speedscope-ubuntu /bin/bash perf-script.sh forks > ../../profiles/linux-perf/forks.linux-perf.txt
 	# docker run --privileged speedscope-ubuntu /bin/bash perf-script.sh with-header > ../../profiles/linux-perf/simple-with-header.linux-perf.txt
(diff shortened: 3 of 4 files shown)


jlfwong:
I'm trying to build the `perf` output for the test file provided in #393.

To do so, in the `/sample/programs/cpp` directory, I run `make linux-perf`, which spins up a docker container, installs linux perf tools, and then executes a script which runs perf on the given program.

Here's what the output looks like at the moment:

```
docker container prune -f
Deleted Containers:
8da1b9e8d59c057cc5697e10a54a1c1a272becdb731f357c7a2b6b91557c4633

Total reclaimed space: 125.8kB
docker image rm speedscope-ubuntu || true
Untagged: speedscope-ubuntu:latest
Deleted: sha256:75ccfbd75d99caa7d5f5074926fd252c1be59938e9ee3bc53b7219d31dcc5e9c
docker build -t speedscope-ubuntu .
[+] Building 0.6s (12/12) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                       0.0s
 => => transferring dockerfile: 347B                                                                                                                                                                                                                                       0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                          0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                            0.0s
 => [internal] load metadata for docker.io/library/ubuntu:bionic                                                                                                                                                                                                           0.5s
 => [internal] load build context                                                                                                                                                                                                                                          0.0s
 => => transferring context: 5.09kB                                                                                                                                                                                                                                        0.0s
 => [1/7] FROM docker.io/library/ubuntu:bionic@sha256:478caf1bec1afd54a58435ec681c8755883b7eb843a8630091890130b15a79af                                                                                                                                                     0.0s
 => CACHED [2/7] RUN apt-get update                                                                                                                                                                                                                                        0.0s
 => CACHED [3/7] RUN apt-get install -y build-essential linux-tools-common linux-tools-4.15.0-30-generic libpthread-stubs0-dev                                                                                                                                             0.0s
 => CACHED [4/7] RUN ln -sf /usr/lib/linux-tools/4.15.0-30-generic/perf /usr/bin/perf                                                                                                                                                                                      0.0s
 => CACHED [5/7] WORKDIR /workdir                                                                                                                                                                                                                                          0.0s
 => CACHED [6/7] ADD . /workdir                                                                                                                                                                                                                                            0.0s
 => CACHED [7/7] RUN make clean simple simple-terminates forks idle                                                                                                                                                                                                        0.0s
 => exporting to image                                                                                                                                                                                                                                                     0.0s
 => => exporting layers                                                                                                                                                                                                                                                    0.0s
 => => writing image sha256:75ccfbd75d99caa7d5f5074926fd252c1be59938e9ee3bc53b7219d31dcc5e9c                                                                                                                                                                               0.0s
 => => naming to docker.io/library/speedscope-ubuntu                                                                                                                                                                                                                       0.0s

Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
docker run --privileged speedscope-ubuntu /bin/bash perf-script.sh idle > ../../profiles/linux-perf/idle.linux-perf.txt
+ [[ idle == \s\y\s\t\e\m\-\w\i\d\e ]]
+ [[ idle == \f\o\r\k\s ]]
+ [[ idle == \i\d\l\e ]]
+ perf record -a -F 999 -g ./idle
Couldn't record kernel reference relocation symbol
Symbol resolution may be skewed if relocation was used (e.g. kexec).
Check /proc/kallsyms permission or run as root.
[ perf record: Woken up 1 times to write data ]
0xe8 [0x6c73]: failed to process type: 1768383842
[ perf record: Captured and wrote 0.116 MB perf.data ]
+ echo 'Done recording'
+ ls -alh perf.data
+ perf script -i perf.data
0xe8 [0x6c73]: failed to process type: 1768383842
make: *** [linux-perf] Error 234
```

The problem line seems to be the following:

```
0xe8 [0x6c73]: failed to process type: 1768383842
```

I have no idea what's causing this, and I'm not sure how to diagnose further.

javierhonduco:
Hi @jlfwong! Found this issue by chance while checking out speed scope and I think this might be happening due to the way the `perf.data` file is being created. As the `perf record` output is being saved there, we aren't only storing the actual binary data produced by perf, but the test program's output as well.

You can see this by running:

```
$ docker run -it --privileged speedscope-ubuntu /bin/bash
root@448b544d0dfd:/workdir# # And now, inside of the container
root@448b544d0dfd:/workdir# perf record -a -F 999 -g ./idle > perf.data
root@448b544d0dfd:/workdir# strings perf.data | grep "beg"
sleep begin
sleep begin
sleep begin
sleep begin
```

which is the output of `./idle` :)

Added a possible way of sorting this out inline

sample/programs/cpp/perf-script.sh
```diff
@@ -14,6 +14,14 @@ if [[ "$1" == "forks" ]]; then
   exit 0
 fi
 
+if [[ "$1" == "idle" ]]; then
+  perf record -a -F 999 -g ./idle > perf.data
```
javierhonduco, line 18:
```suggestion
  perf record -a -F 999 -g --output=perf.data ./idle
```

The same applies below and above. These commands aren't printing anything yet, but perhaps it would be worth adding this flag instead as if in the future they print something, `perf script` and related tooling will break and it will be a bit confusing 😄

jlfwong:
@javierhonduco Ahhh thanks for the catch! I now have it generating a valid profile. Thank you!

sample/programs/cpp/perf-script.sh
jlfwong: I needed to do something very slightly fancier here, because the stdout of `perf-script.sh` is used directly to generate the output files in the `Makefile`, but this `--output` switch was still a helpful key

PR #409, opened 2022-11-03
Support showing pprof lines from the pprof Line object

dalehamel:
The previous behavior was to use the StartLine of a function as the line number to show in speedscope. However, the Line object has more precise line information, and we should only fallback to StartLine if we don't have this more detailed information.

Looking at the [documentation for the pprof proto](https://github.com/google/pprof/tree/main/proto#general-structure-of-a-profile), this is more how it intends to interpret line information:

> location: A unique place in the program, commonly mapped to a single instruction address. It has a unique nonzero id, to be referenced from the samples. It contains source information in the form of lines, and a mapping id that points to a binary.
> function: A program function as defined in the program source. It has a unique nonzero id, referenced from the location lines. It contains a human-readable name for the function (eg a C++ demangled name), a system name (eg a C++ mangled name), the name of the corresponding source file, and other function attributes.

Here is a sample profile that had line-level info on the Line object of the profile:

Before:

Screen Shot 2022-11-03 at 11 17 11 AM

After:

Screen Shot 2022-11-03 at 11 17 22 AM

NB i have never written anything in typscript before, but I tried to follow CONTRIBUTING.md, and run the necessary linter and tests. Note also that the linter changed other files which I didn't modify, so I didn't commit those changes.


dalehamel:
@jlfwong when you have the time to review ❤️

src/import/__snapshots__/pprof.test.ts.snap
```diff
@@ -223,7 +319,11 @@ Object {
       "col": undefined,
       "file": "/usr/local/Cellar/go/1.10.1/libexec/src/runtime/sys_darwin_amd64.s",
       "key": "runtime.usleep:/usr/local/Cellar/go/1.10.1/libexec/src/runtime/sys_darwin_amd64.s:0",
-      "line": 0,
+      "line": Long {
+        "high": 0,
+        "low": 418,
+        "unsigned": false,
+      },
```
jlfwong, line 326: I'm confused about what's happening here. What JavaScript type is this? I would expect this diff to show a change from `"line": 0` to `"line": 418`, rather than what this is. Is this a `BigInt` type?

jlfwong (commented):
Hi!

Thanks for digging into this and documenting your process. Sorry it took so long to get around to looking at this.

Conceptually this seems fine to me, but I have a question inline about the types involved

src/import/__snapshots__/pprof.test.ts.snap
dalehamel: I have no idea what this is, the code was generated when I regenerated the fixtures 🤷‍♂️ 

typescript isn't my forte so any advice would be appreciated.

vasi-stripe: In protobufjs, we [pull in the `long` module](https://github.com/protobufjs/protobuf.js/blob/56b1e64979dae757b67a21d326e16acee39f2267/src/util/minimal.js#L182). And then this apparently causes protobuf decoding to create `Long` objects instead of JS numbers.

For some reason, parcel.js doesn't seem to pull `long` into the web build, and so this only happens in tests, not in the webapp. I'm not familiar enough with JS build systems to understand how to fix that.

src/import/pprof.ts
```diff
@@ -81,7 +81,13 @@ export function importAsPprofProfile(rawProfile: ArrayBuffer): Profile | null {
...
 
     if (lastLine.functionId) {
-      return frameInfoByFunctionID.get(i32(lastLine.functionId)) || null
+      let funcFrame = frameInfoByFunctionID.get(i32(lastLine.functionId))
+      if (lastLine.line && lastLine.line > 0) {
+        if (funcFrame != null) {
+          funcFrame.line = lastLine.line as number
+        }
+      }
+      return funcFrame || null
```
jlfwong, line 90: I took a closer look at this. I'm not sure why browser & node behavior is different, but the type signatures do indicate it's more correct here to handle the case that `lastLine.line` is an instance of `Long` instead of assuming it's `number`. This is also why you needed a cast here. Here's what I think this should be doing instead:

```diff
       let funcFrame = frameInfoByFunctionID.get(i32(lastLine.functionId))
-      if (lastLine.line && lastLine.line > 0) {
-        if (funcFrame != null) {
-          funcFrame.line = lastLine.line as number
-        }
+      const line = lastLine.line instanceof Long ? lastLine.line.toNumber() : lastLine.line
+      if (line && line > 0 && funcFrame != null) {
+        funcFrame.line = line
       }
```

If that looks right to you, can you please update the PR and run `npm run jest -- --updateSnapshot` to update the snapshot? After that, this should be good to land

dalehamel: @jlfwong I tried that, but it is giving me a typescript error I don't understand when I try to run npm jest:

```
  ● Test suite failed to run

    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    src/import/pprof.ts:85:45 - error TS2686: 'Long' refers to a UMD global, but the current file is a module. Consider adding an import instead.

    85       const line = lastLine.line instanceof Long ? lastLine.line.toNumber() : lastLine.line
                                                   ~~~~

```

Seems like some kind of a namespace issue? i guess `Long` isn't available by default to check with `instanceof`?

dalehamel: ok, i was able to make it work by also adding an import:

```
import Long = require("long")
```

I've committed the results and pushed.

src/import/pprof.ts
```diff
@@ -2,6 +2,7 @@ import {perftools} from './profile.proto.js'
 import {FrameInfo, StackListProfileBuilder, Profile} from '../lib/profile'
 import {lastOf} from '../lib/utils'
 import {TimeFormatter, ByteFormatter} from '../lib/value-formatters'
+import Long = require("long")
```
dalehamel, line 5: FYI had to add this to make npm jest work

dalehamel:
Status is that this seems to work, it fixes the previous issue we had with the test fixtures, but now there is an import error that fails tests.

jlfwong:
Okay, thanks! I'll take it from here -- I'll fix this on another branch. Thanks for the followthrough

jlfwong:

dalehamel:
thanks so much for the assistance @jlfwong i really appreciate it! we can get off of our fork now i think ❤️

PR #415, opened 2023-03-03
Fix out of bounds error for pprof default type

dalehamel:
It is possible that the defaultSampleType index is out of bounds. This adds a bounds check to ensure it is a valid index.

### DIFF
--- a/src/import/pprof.ts
+++ b/src/import/pprof.ts
@@ -104,8 +104,11 @@ export function importAsPprofProfile(rawProfile: ArrayBuffer): Profile | null {
   }))
 
   const sampleTypeIndex = protoProfile.defaultSampleType
-    ? +protoProfile.defaultSampleType
+    ? +protoProfile.defaultSampleType > sampleTypes.length - 1
+      ? sampleTypes.length - 1
+      : +protoProfile.defaultSampleType
     : sampleTypes.length - 1
+
   const sampleType = sampleTypes[sampleTypeIndex]
 
   const profileBuilder = new StackListProfileBuilder()


jlfwong:
Hi @dalehamel!

In what circumstances does this happen? It seems to me like this would constitute an invalid profile

I'm not opposed to the change, but want to understand why this happens to figure out e.g. if there's someplace upstream where this needs fixing

vasi-stripe:
We're actually using this field wrong!

It's documented as an index into the string_table, NOT into the repeated sample_types. So if you have a pprof containing both cpu & allocs, and the default type is allocs, it might look like:

```
string_table:
  0: ""
  1: "samples"
  2: "count"
  3: "alloc_objects"
sample_types:
  0:
    type: 1
    unit: 2
  1:
    type: 3
    unit: 2
default_sample_type: 3
```

vasi-stripe:
Here's an example in google's pprof tool, where we're printing the debugging representation of a profile: https://github.com/google/pprof/blob/main/profile/profile.go#L566

Note how it's finding which type is the default by comparing `type.Type == profile.DefaultSampleType`—comparing the indexes into the string table, rather than looking at the index within the sample type array.

vasi-stripe:
While trying to fix this, I'm running into a fun issue! When running in the browser, protobufjs seems to give me a number for DefaultSampleType—but when running in npm, it gives me a Long object.

This changes [what this line does](https://github.com/jlfwong/speedscope/blob/0414c2f617742e7fb0cf31a66ac0f77c2f5c0540/src/import/pprof.ts#L106), making a DefaultSampleType appear truth-y in tests, and thus making the pprof/simple.prof choose sampleTypeIndex = 0 instead of sampleTypeIndex = length - 1 = 1.

I'm not familiar enough with JS compilers to figure out why the `long` module ends up included in the npm build, but not the build parcel creates.

dalehamel:
thanks @vasi-stripe

PR #416, opened 2023-03-03
README: Explain total vs. self time

alexcwatt:
It took me a bit when I first started using Speedscope to understand the "total" vs. "self" distinction, and I've heard the same from others. I thought it might be nice to add a little description here in the README - let me know what you think!


alexcwatt:
@jlfwong I just opened this and #417 and would love any feedback you have. I use Speedscope almost daily, thanks for an awesome tool!

PR #424, opened 2023-06-05
Fix pprof defaultSampleType

vasi-stripe:
Fixes #415

* Interprets pprof's defaultSampleType as an index into the string table [as documented in the proto](https://github.com/jlfwong/speedscope/blob/0414c2f617742e7fb0cf31a66ac0f77c2f5c0540/src/import/profile.proto#L85), not as an index into the sample types repeated-field. This allows parsing to succeed when the string-table index is not a valid sampleTypes index, which is common on allocation profiles.
* Update the pprof snapshot. This is necessarily because we were previously interpreting an empty defaultSampleType as truthy-but-zero when long.js is present, ie: the first sample-type. But Speedscope-in-the-browser doesn't seem to include long.js, so our tests were disagreeing with in-browser behavior. With this PR, that should be fixed.

### DIFF
--- a/src/import/__snapshots__/pprof.test.ts.snap
+++ b/src/import/__snapshots__/pprof.test.ts.snap
@@ -10,7 +10,7 @@ Object {
       "line": 0,
       "name": "runtime.main",
       "selfWeight": 0,
-      "totalWeight": 136,
+      "totalWeight": 1360000000,
     },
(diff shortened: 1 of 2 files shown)


jlfwong:
Thank you!

jlfwong:

PR #425, opened 2023-06-06
Stackprof: weight on-cpu samples by period rather than timestamp delta

manuelfelipe:
This attempts to improve the quality of the on-CPU profiles stackprof provides. Rather than weighing samples by their timestamp deltas, which, in our opinion, are only valid in wall-clock mode, this weighs callchains by:

```
S = number of samples
P = sample period in nanoseconds

W = S * P
```

The difference after this change is quite substantial, specially in profiles that previously were showing up with heavy IO frames:  

* Total profile weight is almost down by 90%, which actually makes sense for an on-CPU profile if the app is relatively idle
* Certain callchains that blocked in syscalls / IO are now much lower weight. This was what I was expecting to find.
Here is an example of the latter point.

In delta mode, we see an io select taking a long time, it is a significant portion of the profile:

236936508-709bee01-d616-4246-ba74-ab004331dcd3

But in period scaling mode, it is only a couple of sample periods ultimately:

236936693-9d44304e-a1c2-4906-b3c8-50e19e6f9f27


jlfwong:
Hi @manuelfelipe!

This seems reasonable to me! Can you please include tests which would catch changes to this behaviour were it to break. I think this would mean including a new profile with the `cpu` mode set in https://github.com/jlfwong/speedscope/blob/main/src/import/stackprof.test.ts

manuelfelipe:
thanks @jlfwong. Rebased base on latest changes on main, and re-ran `run jest -- -u` in https://github.com/jlfwong/speedscope/pull/425/commits/2b92d51817072d50d69a5827adce7061f14b581e to try and get CI happy. 


Not sure I fully understand what is the actual thing failing failure in https://github.com/jlfwong/speedscope/actions/runs/5201005116/jobs/9543115054.  Locally tests are passing, so hoping the thing that was causing the issue are the outdated snapshots.

dalehamel:
FYI @jlfwong I enabled actions on my fork, Manuel pushed to it, and it passes there https://github.com/dalehamel/speedscope/actions/runs/5280170594 (since we cannot run workflows here, this seemed the most expedient way).

jlfwong:
Thanks! This will go out with the next release

jlfwong:

Issue #427, opened 2023-06-17
Support for Papyrus profiling


xieve:
I've hacked together a parser/converter from Bethesda's proprietary Papyrus profiling format to speedscope JSON. There is no documentation of the format that I know of, but it isn't complex, parsing & converting takes around 100 LOC. It's a very niche format, but if you're interested, I could implement it natively into speedscope in a PR. Otherwise, I'd probably self-host a fork for the community.

So far, I've tested it with Skyrim, but FO4 profiles should work too.

jlfwong:
Hi @xieve!

Happy to take it provided that...
1. It doesn't require bringing in any large dependencies or cause performance problems elsewhere (by e.g. being expensive to detect what kind of profile it is)
2. You're comfortable being a point of contact for changes needed to it
3. You're able to write tests for it to ensure it doesn't break over time

xieve:
> 1. It doesn't require bringing in any large dependencies or cause performance problems elsewhere (by e.g. being expensive to detect what kind of profile it is)

The file ending is `.log`, which is not used yet and unlikely to be used elsewhere. The contents are plain text and look like this:
```
Script_abc_runner_quest log opened (PC)
21171:PUSH:3053:1:abcRunnerQuest (24021278):ssx_runner_quest..iAmInitialized
21171:POP:3053:1:abcRunnerQuest (24021278):ssx_runner_quest..iAmInitialized
21171:PUSH:3053:1:abcRunnerQuest (24021278):ssx_runner_quest..isInitInProgress
21171:POP:3053:1:abcRunnerQuest (24021278):ssx_runner_quest..isInitInProgress
```
We can easily detect that header.
> 2. You're comfortable being a point of contact for changes needed to it

Absolutely.

> 3. You're able to write tests for it to ensure it doesn't break over time

I don't have any experience writing tests so far, but I'm familiar with the concept. Can you give me a few pointers to documentation?

Another issue to take note of: This profiler tends to output "malformed" data. Here are two examples I encountered while writing the parser:
```
at 1 open a
at 2 open b
at 3 close a
at 3 close b
```
```
at 1 close a
at 2 open b
at 3 close b
```

I wrote a heuristic algorithm that fixes these and possibly even worse cases. Alternatively, I could detect those cases and throw them out.

PR #428, opened 2023-06-18
Added support for Papyrus profiles (#427)

xieve:
Here's what I have so far. There are still a couple of issues:

- I couldn't get the tests to run (`sh: 1: ./scripts/test-setup.sh: not found`)
- I've specified `Profile.totalWeight` but they still always start at 0
- I'm not sure why `Profile.stack` and `Profile.frames` are protected, this way I have to re-implement those


xieve:
Everything works now. As I said, implementation could be more ideal if I could access `Profile.stack` and `Profile.frames`, but apart from that, I'd say this is mergeable.

xieve:
I've noticed one more thing: In the time order view, you can't see the things that take under 1ms, because they count as 0ms, which makes them completely invisible. Calling those functions still takes time, though, as the calls sometimes can't execute instantly and are thus put on the queue instead. I will try to implement queues as frames that contain the function calls.

src/import/papyrus.ts
```diff
@@ -0,0 +1,129 @@
...
+import assert from 'assert'
+import {TimeFormatter} from '../lib/value-formatters'
+import {TextFileContent} from './utils'
+
+/**
+ * Returns the last element of `arr`
+ */
+function lastElement<T>(arr: T[]): T {
+  return arr.slice(-1)[0]
+}
```
jlfwong, line 25: This already exists in `lib/util.ts`, but returns `T | null`, as it should

src/import/papyrus.ts
```diff
@@ -0,0 +1,129 @@
...
+// directly before their corresponding "PUSH" event. Other "QUEUE_PUSH" (and all "QUEUE_POP") events come, as far as I
+// can tell, from Events and are thus asynchronous. They are ignored.
+
+// Stack profiling also puts a START operation at the top and a STOP operation at the bottom of the file. These are
+// completely useless to us.
+
+// Stack profiling also logs the Form a method is run on. For those that are not familiar with
+// Papyrus terminology, a "Form" is an instance of a type defined by a script. E.g. a quest is a form that extends
+// the "Quest" script, and thus it has certain methods, like "CompleteQuest()". This information would be useful
+// for Debugging, but for profiling, it would hinder meaningful output in left heavy mode.
```
jlfwong, line 12: Thank you for including a top-of-file comment like this! One small nice addition would be to explain what Papyrus is, since it's relatively obscure

src/import/papyrus.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export function importFromPapyrus(papyrusProfile: TextFileContent): Profile {
+  const profile = new CallTreeProfileBuilder()
+  profile.setValueFormatter(new TimeFormatter('milliseconds'))
+
+  const papyrusProfileLines = papyrusProfile
+    .splitLines()
+    .filter(line => line.match(/^$|^Log closed$|log opened/) === null)
+  const startValue = Number(papyrusProfileLines[0].split(':')[0])
+  const endValue = Number(lastElement(papyrusProfileLines).split(':')[0]) - startValue
```
jlfwong, line 35: I'm guessing you're expecting both of these to be integers, right? Any reason not to use `parseInt` here instead?

src/import/papyrus.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  const nameSet: KeyedSet<Frame> = new KeyedSet<Frame>()
```
jlfwong, line 37: This is fine as-is, but you don't need the type annotation. `const nameSet = new KeyedSet<Frame>()` is fine

src/import/papyrus.ts
```diff
@@ -0,0 +1,129 @@
...
+// completely useless to us.
+
+// Stack profiling also logs the Form a method is run on. For those that are not familiar with
+// Papyrus terminology, a "Form" is an instance of a type defined by a script. E.g. a quest is a form that extends
+// the "Quest" script, and thus it has certain methods, like "CompleteQuest()". This information would be useful
+// for Debugging, but for profiling, it would hinder meaningful output in left heavy mode.
+
+import {CallTreeProfileBuilder, Frame, Profile} from '../lib/profile'
+import {KeyedSet} from '../lib/utils'
+import assert from 'assert'
```
jlfwong, line 16: I don't think I use this anywhere else in the codebase, and it weirds me out having dependencies that are part of the nodejs standard library since they get different implementations in-browser.

It looks like it's only used in a single place -- can you please update this to just throw an error in that case?

jlfwong (changes requested):
Thank you!

A few small things that need revision, but this should be good to land afterwards

xieve:
Implemented all suggested changes :)

xieve:
I had to change more than I anticipated, everything seems to be working as intended now.

jlfwong:
Hey @xieve -- just opened a test run, and it looks like CI is failing at the moment. You can see the issue here: https://github.com/jlfwong/speedscope/actions/runs/5317035811/jobs/9678064431?pr=428

It seems like the snapshots don't match -- can you take a look, update the snapshots, and make sure the changes are as expected?

xieve:
Damn, I forgot to commit the updated snapshots.

PR #431, opened 2023-06-23
Fix trimTextMid

### LINKED ISSUE #411: Frame label truncates to second last character instead of last
image

From importing the sample file `sample/profiles/Chrome/70/worker.json`

jlfwong:
There was a subtle bug in `trimTextMid` caused by calling substring methods with non-integer values. This happens because `findValueBisect` returns non-integer values, and there was no special handling of this.

The bug results in strings cutting off many of the last few characters in a string, rather than always displaying it when possible.

Before:
image

After:
image

Fixes #411

### DIFF
--- a/src/lib/text-utils.ts
+++ b/src/lib/text-utils.ts
@@ -43,8 +43,8 @@ export function buildTrimmedText(text: string, length: number): TrimmedTextResul
 
   let prefixLength = Math.floor(length / 2)
   const suffixLength = length - prefixLength - 1
-  const prefix = text.substr(0, prefixLength)
-  const suffix = text.substr(text.length - suffixLength, suffixLength)
+  const prefix = text.substring(0, prefixLength)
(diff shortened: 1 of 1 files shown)


src/lib/text-utils.ts
```diff
@@ -43,8 +43,8 @@ export function buildTrimmedText(text: string, length: number): TrimmedTextResul
 
   let prefixLength = Math.floor(length / 2)
   const suffixLength = length - prefixLength - 1
-  const prefix = text.substr(0, prefixLength)
-  const suffix = text.substr(text.length - suffixLength, suffixLength)
+  const prefix = text.substring(0, prefixLength)
+  const suffix = text.substring(text.length - suffixLength, text.length)
```
jlfwong, line 47: `substr` is now deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

Issue #433, opened 2023-06-27
Importing a large linux perf file will crash the application


Goose97:
Hi, thank you for your work on this awesome project.

Back to the story, I was importing a large file (linux perf tool format), around ~1.3 GB and it crashed the application. The root cause is [this line](https://github.com/jlfwong/speedscope/blob/bb063e49e305210316681369d3d6eb8e779cd58b/src/import/linux-tools-perf.ts#L103). When we're splitting the file into blocks, a block size can exceeds V8 strings limit. The situation is similar to #385.

My plan is:
1. change `splitBlocks` signature to
```typescript
splitBlocks(contents: TextFileContent): TextFileContent[]
```
2. pending will now use `StringBackedTextFileContent | BufferBackedTextFileContent`. It will initially use `StringBackedTextFileContent` for performance then fallback to use `BufferBackedTextFileContent` if the block size exceeds the limit. Something like this
```typescript
try {
  pending += line
} catch (error) {
  if (check(error)) {
    // convert to BufferBackedTextFileContent and continue
  }
}
```
3. To do this, we might need to add some methods to TextFileContent interface:
- isEmpty(): to replace [pending.length > 0](https://github.com/jlfwong/speedscope/blob/bb063e49e305210316681369d3d6eb8e779cd58b/src/import/linux-tools-perf.ts#L106) 
- append(): to replace string concatenation

Another approach that I'm considering is to implement a generic `splitBy(delimiter: string)` for TextFileContent. I haven't thought it through yet but it seems like a non-trivial thing.

The file is fairly large file so I won't attach it here. I still keep it in my computer in case you want to have a look.

jlfwong:
Hey @Goose97!

Thanks for the very clear description of the problem you'er trying to solve, and the links to the specific parts of the code!

If I'm understanding the problem correctly, I think there's a simpler solution that's probably more performant.

The `splitBlocks` function doesn't really need to exist at all. It was just a simple way to reason about the problem.

Instead, I think we can change the importer to just operate on the output of `splitLines()` directly without using `\n\n` delimited blocks as an intermediate. This also has the benefit of not doubling the memory requirements.

If we do your proposed change, I *think* we end up with the 1.3GB in memory three times: once as a buffer, once as the return value of `.splitLines()` and once as the `TextFileContent[]`. If we skip the `splitBlocks`, then we just have the first two.

[...]

PR #434, opened 2023-06-28
Improve splitLines: return iterator instead

Goose97:
The current behavior of splitLines is to eagerly split all the lines and return an array of strings.

This PR improves this by returning an iterator instead, which will emit lines. This lets callers decide how to best use the splitLines function (i.e. lazily enumerate over lines)

Relates to #433


jlfwong:
Nice! Thank you

PR #437, opened 2023-06-29
Improve profile builder performance

Goose97:
Follow up on this PR #435. 

Currently, it took roughly 22 seconds to load my 1.3GB file. After inspecting the profiler, there's a large chunk of time spending in Frame.getOrInsert. I figure we can reduce the number of invocations by half. It reduces the load time to roughly 18 seconds.I also tested with a smaller file (~350MB), and it show similar gains, about 15-20%

I know this is not a proper way to perform benchmarking, and also I'm clueless about the impact on other file formats/sizes. So I'll leave the judgement to you

### DIFF
--- a/src/lib/profile.ts
+++ b/src/lib/profile.ts
@@ -439,14 +439,13 @@ export class Profile {
 }
 
 export class StackListProfileBuilder extends Profile {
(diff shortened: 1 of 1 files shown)


jlfwong:
Wonderful! Nice tactical performance fix

Issue #426, opened 2023-06-15
Maximum call stack size exceeded


zlsun:
Similar to #327 
```
speedscope.c62e67c1.js:183 Uncaught RangeError: Maximum call stack size exceeded
    at maybeResize (speedscope.c62e67c1.js:183:2113)
    at onBeforeFrame (speedscope.c62e67c1.js:30:725)
    at speedscope.c62e67c1.js:14:6435
    at Set.forEach (<anonymous>)
    at a.resize (speedscope.c62e67c1.js:14:6424)
    at maybeResize (speedscope.c62e67c1.js:183:2186)
    at onBeforeFrame (speedscope.c62e67c1.js:30:725)
    at speedscope.c62e67c1.js:14:6435
    at Set.forEach (<anonymous>)
    at a.resize (speedscope.c62e67c1.js:14:6424)
```

I try to debug it, and find that it was caused by float point precision

![image](https://github.com/jlfwong/speedscope/assets/6016980/768d1aae-613d-4437-b4e1-2a6f70051c64)

```
n.width: 2066.004150390625
i: 2066.0166015625
n.height: 494.50836181640625
r: 494.5209655761719
```

System Info:
- System: Windows 10
- Monitor resolution: 2560 x 1440
- System Custom scaling: 124%

PR #438, opened 2023-07-14
Fix maximum call stack size exceeded in some case

Camork:
I found some stackoverflow bugs when import a customized format with a  very deep recursion. There are another similar [report ](https://github.com/jlfwong/speedscope/issues/426), but  I can only fix my cases.

### DIFF
--- a/src/lib/profile.ts
+++ b/src/lib/profile.ts
@@ -188,8 +188,15 @@ export class Profile {
   // as immutable.
   protected sortGroupedCallTree() {
     function visit(node: CallTreeNode) {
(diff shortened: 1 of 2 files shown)


jlfwong:
Hi @Camork!

Can you share more about how the stack overflow was was being produced in the first place? Is this a profile you produced where the actual stack depth of the profile itself was in the hundred to thousands?

Camork:
We have implement a format in our own , it records the function call, and use the `CallTreeProfileBuilder` to build the `ProfileGroup` .  The format composed by the'call' and 'return' like function.

> Is this a profile you produced where the actual stack depth of the profile itself was in the hundred to thousands?

Yes, we have record a application which do recursion call about thirty thousand mostly.

PR #450, opened 2023-12-15
Fix bug in selectQueueToTakeFromNext for trace profiles

zacharyfmarion:
I have been taking a lot of profiles using the Hermes profiler, but I noticed that they sometimes to not show up properly. After debugging what exactly was going on, I realized it was because the logic in `selectQueueToTakeFromNext` only checks for name, instead of the key for the event. I had a bunch of events with the name `anonymous` that were getting improperly exited before they should have been due to this logic. 

This fix makes the code more robust if there are added "args" which differentiate an event from another (as is the case in Hermes profiles), however it would still be an issue if they key just defaults to the name.

Example profile before:

Screenshot 2023-12-15 at 12 54 04 AM

What it should look like (in Perfetto):
Screenshot 2023-12-15 at 8 51 38 AM

After the fix:

Screenshot 2023-12-15 at 12 54 29 AM


zacharyfmarion:
@jlfwong can you take a look?

jlfwong:
Hey @zacharyfmarion! Thanks for the investigation and the patch!

An example trace sent privately would be appreciated (you can send to jamie.lf.wong at gmail), but even better would be a dramatically reduced test file that minimally reproduces the issue.

src/import/trace-event.ts
```diff
@@ -100,15 +112,21 @@ function selectQueueToTakeFromNext(
   if (bts < ets) return 'B'
   if (ets < bts) return 'E'
 
-  // If we got here, the 'B' event queue and the 'E' event queue have events at
-  // the front with equal timestamps.
+  // If the keys are the same then we need to check whether the number of begin events
+  // with that key are the same as the number of end events with that key for the given timestamp.
+  // - If the number of begin and end events are the same, then we have a zero duration event,
+  // and we should process the begin first
+  // - In all other cases we should process the end event first
```
jlfwong, line 119: I'm having trouble following this logic. If you had, conceptually:

```
B funcX ts=1
B funcY ts=1
E funcX ts=1
E funcY ts=2
```

Then the correct processing order is:

- B funcX ts=1
- E funcX ts=1 (a zero-duration event)
- B funcY ts=1
- E funcY ts=2

But if I'm reading the description here correctly, this code would process in the following order, because the number of begin events and end events at the given timestamp are different:

- E funcX ts=1
- B funcX ts=1
- B funcY ts=1
- E funcY ts=2

Am I misunderstanding?

zacharyfmarion: Sorry I didn't realized I pushed this up - this is not quite right. For now, I'm going to push up the original change, which just uses keys to avoid this logic for my case and is better. I'll make a separate PR to fix the general case, which is a bit more complicated and I need to think about some more

src/import/trace-event.ts
```diff
@@ -100,15 +100,15 @@ function selectQueueToTakeFromNext(
   if (bts < ets) return 'B'
   if (ets < bts) return 'E'
 
-  // If we got here, the 'B' event queue and the 'E' event queue have events at
+   // If we got here, the 'B' event queue and the 'E' event queue have events at
```
jlfwong, line 103: Super tiny nit: can you revert this whitespace change? it looks like it was accidental

zacharyfmarion: Yep - also I ran `prettier` and a bunch of files changed so I didn't commit that, I can push up a follow up commit

Issue #201, opened 2019-01-25
TraceEvent scopes should be grouped by scope name only


uucidl:
I tested speedscope with TraceEvent files (generated with https://github.com/uucidl/uu.spdr)

When the scopes contain some arguments, speedscope helpfully shows it right there in the chart:

`Foo::Bar() {"Baz": 6}`

However it appears that aggregates/groups of scopes are using the scope name + args instead of
just the scope name. For the type of performance analysis I do, arguments are metadata around
the scope, but shouldn't be considered different scopes. 

(Note this was tested with the self-contained release of speedscope: speedscope-1.4.1)

jlfwong:
Hi @uucidl!

Yeah, I figured that would be contentious when I was implementing it.

In the profiles I was testing on, like the ones generated by @aras_p's wonderful clang profiler (https://aras-p.info/blog/2019/01/16/time-trace-timeline-flame-chart-profiler-for-Clang/), metadata about individual files is stored in the details.

In the case of the profiles in that blogpost, if aggregation was done just on the basis of the name, then it would just say something like "80% is spent in `SourceFile`", which would be pretty unhelpful.

The option here that's most appealing to me is to continue with the default behavior that speedscope currently has, but provide a way of specifying in trace files that you don't want to include args in aggregration. This could be done by using the `{traceEvents: [...], ...}` format specified in the Trace Events spec. That might look something like this:

```
{
  traceEvents: [{...}, ...],
  speedscope: {
    ignoreArgs: true
  }
}
```

By far the easiest version of this would be to have it no longer display the args in the call frames at all then. Having call frames display different things in the time-ordered view but still be aggregated together in the left heavy and sandwich views isn't something speedscope supports at the moment. It's possible to build it, but it might be a gnarly data modeling change (I haven't thought too hard about it).

speedscope in general doesn't have any way to display arbitrary metadata at the moment, so the imports of those profiles would've been totally useless if that metadata wasn't included somehow.

Would providing a way to ignore `args` completely satisfy your use case? Another way of phrasing this is "if you ran a preprocessing script on your profile that stripped out all `args` before importing into speedscope, would that meet your needs"?

uucidl:
I think I could live with the current behavior for now until speedscope eventually develops the ability to attach, visualize and analyze metadata on scopes in its other formats. I.e. if it includes them as a first class concept.

[...]

PR #451, opened 2023-12-16
Default frame name to the name of the function call

zacharyfmarion:
Because the displayed frame name is the entire key, the display name becomes very hard to parse if there are args that get stringified into the key (this is the case for Hermes profiles). Defaults to just showing the name for trace profiles.

| Before | After |
| ------ | ----- |
| Before 1 | After 1 |
| Before 2 | After 2 |
| Before 3 | After 3 |


zacharyfmarion:
@jlfwong happy to iterate on this, no rush but want to make sure you agree this is a good approach. Also would it be possible to release a new version of speedscope with this commit and my previous one one they are good to go? Would like my company's engineers to switch over from using Perfetto since this UI is way better.

jlfwong:
Hey -- unfortunately if this information is hidden, then the args information isn't visible anywhere at all in speedscope. There's no place to display metadata unrelated to the file & line information.

Others have, in the past, indicated that the display of this information is helpful. For example: https://github.com/jlfwong/speedscope/issues/201#issuecomment-458019528

zacharyfmarion:
@jlfwong gotcha, makes sense, for now I can just fork. Do you agree that the profile afterward is easier to read though? Maybe we can iterate on an approach. What if we just showed all the args in the tooltip and footer panel?

jlfwong:
Hm, I don't have a strong enough intuition for how people use args, especially given that the chrome trace format is spit out by a really wide variety of profiling tools. I have general gripes with the chrome trace format because it leaves so much behavior unspecified.

I also don't know how anything about the Hermes ecosystem, but is it possible to submit a patch to export directly into speedscope's own format to avoid a bunch of these annoying ambiguities? Or, alternatively, I'd accept a patch which ingests Hermes own internal profiling format (which would hopefully obviate the need for using https://github.com/react-native-community/hermes-profile-transformer?tab=readme-ov-file, which I'm assuming you're using?)

zacharyfmarion:
@jlfwong thanks for the response - I have opened a [PR](https://github.com/jlfwong/speedscope/pull/453) to support the hermes format in speedscope, since I think it is good to have. I still think that the ideal case is that `hermes-profile-transformer` exports into a format that perfetto, chrome devtools, and speedscope can all read. I agree the spec is kind of bad but given that it is what we already have and widely supported, are you strongly opposed to me just check if all the args match a hermes profile, and having custom logic to format Hermes profiles well if that is the case? I talk a bit more about it in my hermes support PR.

PR #453, opened 2023-12-21
Support the chrome JSON trace format (allows viewing of hermes traces)

zacharyfmarion:
Add support for the chrome [JSON object trace format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.q8di1j2nawlp). I have tested this with hermes profiles, which spit out a version of the spec with additional information in the stack frame. Mainly inspiration was taken from two sources:

* [React native profiling docs](https://reactnative.dev/docs/profile-hermes#how-does-the-hermes-profile-transformer-work)
* [hermes-profile-transformer package](https://github.com/react-native-community/hermes-profile-transformer)

[Example profiles](https://drive.google.com/drive/u/0/folders/1--E9sLk-4Iti_ddRozXkrYw4VH8auEPa)

Specifically I pulled some code from [this file](https://github.com/react-native-community/hermes-profile-transformer/blob/master/src/profiler/cpuProfilerModel.ts). While this works great for raw profiles, the `hermes-profile-transformer` package also applies sourcemaps, which is something that I would like to by applied for hermes profiles as well. It's a little awkward because in the hermes profile file format, there isn't actually a field from the file path where a given function in a frame originated - instead it is in parentheses after the function name. Also I will need to update the implementations in three places:

1. https://github.com/react-native-community/hermes-profile-transformer
2. https://github.com/react-native-community/cli
3. Bump version in [react](https://github.com/facebook/react-native/blob/main/packages/react-native/package.json#L98)

Which is a little overwhelming. Given time constraints I might not get a chance to land all of this for a while. I think in the shorter term it would be nice to just add some specific support for hermes-generated trace profiles in the trace-event file. Discussion on this is [here](https://github.com/jlfwong/speedscope/pull/451)

## Example

| Trace event format | Trace event object format |
|------- | ------- |
|  Screenshot 2023-12-20 at 10 21 29 PM | Screenshot 2023-12-20 at 10 21 55 PM |


src/import/index.ts
```diff
@@ -189,8 +194,8 @@ async function _importProfileGroup(dataSource: ProfileDataSource): Promise<Profi
       return toGroup(importFromInstrumentsDeepCopy(contents))
     }
 
-    if (/^(Stack_|Script_|Obj_)\S+ log opened \(PC\)\n/.exec(contents.firstChunk())){
-      console.log("Importing as Papyrus profile")
+    if (/^(Stack_|Script_|Obj_)\S+ log opened \(PC\)\n/.exec(contents.firstChunk())) {
```
zacharyfmarion, line 197: All changes unrelated to hermes are from running `npm run prettier`. Might be worth adding `npm run prettier --check` to the ci.yml so PRs fail if committers have not run the formatter

zacharyfmarion:
Actually after looking at this some more it looks like spec is actually part of the [chrome trace event spec](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#bookmark=id.end1pppx3rl)? But it doesn't currently seem to be implemented...at least hermes profiles did not work before. @jlfwong should i name this something more generic?

jlfwong:
Going to continue discussion of both https://github.com/jlfwong/speedscope/pull/451#issuecomment-1865423405 and this PR here, since they're intertwined.

Given that this is an extension of the chrome trace event spec, this should IMO live inside of `import/trace-event.ts`. That file already contains handling of both a `TraceEvent[]` format and a `{traceEvents: TraceEvent[]}` format, so this seems like a fairly natural extension. And that unfortunately means it similarly should be generic w.r.t. handling of `args`.

That said, if there are highly reliable ways to see from analysis of a profile that it was specifically generated by Hermes, whether it be a `{traceEvents, stackFrames, samples}` profile as in this PR, or one of the existing supported trace event formats, I'm open to accepting a patch to `trace-event.ts` which ignores args in that case because they're known to be useless when coming from Hermes.

This could be through structural analysis of the JSON, or it could be from filename matching if Hermes spits out files with a specific naming convention.

Does that sound tenable?

zacharyfmarion:
@jlfwong Okay sounds good - I will refactor trace-event to support the json object format - are you okay with me keeping the main json import code in a separate file? The trace-event file is already pretty large. Not sure if you're opposed to me creating a folder in the imports directory called `traceEvent` that has a `types` file and then two files, one for each import type. I can put it all in one file if you prefer.

Re. the hermes format - I should be able to reliably detect if the args match what hermes exports - thanks for understanding. I will do that after this lands and we can iterate. I definitely don't want to merge anything you're not comfortable with.

src/import/trace-event-json.ts
```diff
@@ -0,0 +1,229 @@
...
+   */
+  function getActiveNodeIds(nodeId: number): number[] {
+    const activeNodeIds = activeNodeArraysById.get(nodeId)
+    if (!activeNodeIds) throw new Error(`No such node ID ${nodeId}`)
+    return activeNodeIds
+  }
+
+  // We need to leave frames in the same order that we start them, so we keep a stack
+  // of frames that are currently open
+  const frameStack: StackFrame[] = []
```
zacharyfmarion, line 95: There is an opportunity to de-duplicate some of this logic i can look into in a bit

src/import/trace-event.test.ts
```diff
@@ -122,3 +122,7 @@ test('importTraceEvents invalid x nesting', async () => {
 test('importTraceEvents event reordering name match', async () => {
   await checkProfileSnapshot('./sample/profiles/trace-event/event-reordering-name-match.json')
 })
+
+test('importTraceEvents simple json profile', async () => {
```
zacharyfmarion, line 126: I can add a test for supporting multiple threads but I don't know where to get it - @jlfwong do you know where to get one?

zacharyfmarion:
Went ahead and pushed it up, can clean up a bit later

src/import/trace-event.test.ts
zacharyfmarion: Added test [here](a565f1d4eb73ac1e92e4cf8275da4bf48dda8c26)

src/import/trace-event-json.ts
```diff
@@ -0,0 +1,229 @@
...
+    frameStack.push(frame)
+    profileBuilder.enterFrame(frameInfoForEvent(frame), timestamp)
+  }
+
+  /**
+   * Attempt to leave a frame. First we check if the frame matches what we expect to be the
+   * next thing to leave (top of the stack). If this is not the case we warn, and then leave
+   * the frame at the top of the stack
+   */
+  function tryToLeaveFrame(frame: StackFrame, timestamp: number) {
```
zacharyfmarion, line 111: There is some duplication here but I'm not sure if the juice is worth the squeeze...happy to refactor if you think otherwise

jlfwong:
@zacharyfmarion thanks for being so responsive to feedback! is this ready to review now, or should I hold off for further changes?

zacharyfmarion:
@jlfwong ready for review!

src/import/trace-event-json.ts
```diff
@@ -0,0 +1,229 @@
...
+  const map: Map<number, number[]> = new Map<number, number[]>()
+
+  // Given a nodeId, `getActiveNodes` gets all the parent nodes in reversed call order
+  const getActiveNodes = (id: number): number[] => {
+    if (map.has(id)) return map.get(id) || []
+
+    const node = profile.stackFrames[id]
+    if (!node) throw new Error(`No such node ${id}`)
+    if (node.parent) {
+      const array = getActiveNodes(node.parent).concat([id])
```
jlfwong, line 31: This gives me mild pause. This is cloning the array for every child, which seems like it could end up being a ton of memory. If you have a call-stack that's N frames deep, this will allocate O(N^2) memory because the root node will be in N arrays, the one below it in N-1 arrays, etc.

I think this is irrelevant, however, if you switch to using `StackListProfileBuilder` instead though

jlfwong (commented):
Okay, digging in! Thanks for getting this working, and I hope you'll bear with me for a fairly thorough review since this is changing control flow for codepaths I update semi-frequently.

First, and probably most importantly, I think you want to use `StackListProfileBuilder` instead of `CallTreeProfileBuilder` for this. I think that will obviate the need for more than half the code in `trace-event-json.ts`.

Second, I have some really high-level feedback on control flow and type structuring. There are two principles that I try to uphold in the codebase that this is deviating from:

1. Whenever it isn't performance prohibitive, default to pure functions. The clearest deviation from that in this PR is the `buildFunction` which mutates a `ProfileBuilder` in place.

2. Whenever possible, I try to keep types semantically meaningful, rather than just balls of fields that are incidentally related because of the particular control flow of the code. One indication of when this is happening is that some of the fields are only accessed in a single function somewhere because it was a convenient way of threading them through the control logic. The clearest deviation of this is the existence of `ProfileBuilderInfo`, where, for example, `pid` and `tid` are only accessed in `jsonObjectTraceBuilder`, and `importableEvents` is only accessed in `constructProfileFromTraceEvents`. Once you remove those fields which are only accessed in one place, the only thing left is the `ProfileBuilder`. Usually this kind of code smell can be resolved by deleting the type and just threading the relevant variables where they actually need to be.

I think in this particular case, both of these problems can be resolved by the extraction of a few helper functions (e.g. a `getNameForProfile(pid, tid, processNamesByPid, threadNamesByTid): string`) function, and then dropping the `buildFunction` callback altogether.

Instead, similar to the pattern that pre-dates this PR of having an `eventListToProfileGroup` function which returns a `ProfileGroup`, you could have an `sampleListToProfileGroup` which is still a pure function. It will create slightly more code duplication, but I think the clarity of control flow is worth it.

zacharyfmarion:
@jlfwong Thank you for the thorough review - I will likely take a look at this after Christmas - I totally missed `StackListProfileBuilder` 🤦 , will read through that code and update!

src/import/trace-event.ts
```diff
@@ -65,16 +66,47 @@ interface XTraceEvent extends TraceEvent {
...
+  ts: string
+  pid: number
+  tid: string
+  weight: string
+  // Will refer to an element in the stackFrames object
+  sf: number
+  stackFrameData?: StackFrame
+}
+
+export interface TraceEventJsonObject {
```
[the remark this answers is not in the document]
zacharyfmarion: `TraceEventWithSamples` to me sounds like it s a single `TraceEvent` with additional fields, instead of an object that contains an array of traceEvents, along with samples. How about `ChromeTraceWithSamples`?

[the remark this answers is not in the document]
jlfwong: Ah, naming things is hard. `ChromeTraceWithSamples` sounds to me like it's specific to Chrome, and there's a whole separate importer in speedscope for Chrome profiles specifically (since they have their own format for CPU profiles that's totally unrelated to trace event -- the JavaScript profiler and the full-browser profiler are totally separate.

But I hear you that `TraceEventWithSamples` is also not a good name. `TraceEventsObjectWithSamples`?

src/import/trace-event.ts
```diff
@@ -365,50 +454,160 @@ function eventListToProfileGroup(events: TraceEvent[]): ProfileGroup {
...
+    if (!samplesForPidTid) {
+      throw new Error(`Could not find samples for key: ${samplesForPidTid}`)
+    }
+
+    if (samplesForPidTid.length === 0) {
+      return
+    }
+
+    profilePairs.push([pidTidKey, constructProfileFromSampleList(contents, samplesForPidTid, name)])
   })
```
zacharyfmarion, line 611: @jlfwong not sure how you feel about this duplication...I can abstract some stuff away or just keep it like this depending on what you think

jlfwong:
@zacharyfmarion looks like you couldn't wait until after Christmas 😄

High level this looks great! Thanks for taking all the feedback in stride. It looks like there are tests failing now, so it seems like the refactor changed the output. Let me know when that's resolved and this is ready for another look

zacharyfmarion:
@jlfwong hahaha yep got antsy! Tests were correctly failing as I thought "weight" represented the time diff between the previous frame and current frame, instead of the current frame and the next one. Updated and it looks like the profiles look the same as before the refactor. Should be ready for another round of feedback whenever you get a chance!

src/import/index.ts
jlfwong: This was already part of the repo (though using an eslint plugin instead of `prettier --check` for reasons I don't remember)

But I broke it 3 years ago! PR to fix that is here: #454 

I'm going to wait until this merges though to avoid you dealing with annoying merge conflicts

src/import/trace-event.ts
```diff
@@ -65,16 +71,53 @@ interface XTraceEvent extends TraceEvent {
...
+  weight: string
+  // Will refer to an element in the stackFrames object
+  sf: number
+  stackFrameData?: StackFrame
+}
+
+interface TraceWithSamples {
+  traceEvents: TraceEvent[]
+  samples: Sample[]
+  stackFrames: {[key in string]: StackFrame}
```
jlfwong, line 100: I've never seen `[key in string]` used here -- this is the same as `[key: string]: StackFrame`, right?

The `in` syntax is just for mapped types AFAIK?

src/import/trace-event.ts
```diff
@@ -365,50 +454,166 @@ function eventListToProfileGroup(events: TraceEvent[]): ProfileGroup {
...
+  while (parent) {
+    const frame: StackFrame = stackFrames[parent]
+
+    if (!frame) {
+      throw new Error(`Could not find frame for id ${parent}`)
+    }
+
+    frames.push(frameInfoForSampleFrame(frame))
+    parent = frame.parent
+  }
```
jlfwong, line 534: Technically there could be a cycle here that would cause the browser to hang, but I guess we'll worry about that if someone finds a real profile in the wild that does that. It wouldn't be a valid profile, but it's certainly possible to construct a file that would do this

src/import/trace-event.ts
```diff
@@ -365,50 +454,166 @@ function eventListToProfileGroup(events: TraceEvent[]): ProfileGroup {
...
+  profileBuilder.setName(name)
+
+  const timeDeltas = getTimeDeltasForSamples(samples)
+
+  samples.forEach((sample, index) => {
+    const timeDelta = timeDeltas[index]
+    const activeFrames = getActiveFramesForSample(contents.stackFrames, sample.sf)
+
+    profileBuilder.appendSampleWithWeight(activeFrames, timeDelta)
+  })
```
jlfwong, line 556: Haha yeah that looks a lot simpler

src/import/trace-event.ts
```diff
@@ -365,50 +454,166 @@ function eventListToProfileGroup(events: TraceEvent[]): ProfileGroup {
...
+  }
+
+  return frames.reverse()
+}
+
+function constructProfileFromSampleList(
+  contents: TraceWithSamples,
+  samples: Sample[],
+  name: string,
+) {
```
jlfwong, line 543: Add a return type to the function signature please

I have the eslint rule dispabled to always enforce return types being required because sometimes it's absurdly obvious or annoying to do, but here not's not trivially obvious.

Also for consistency, please rename to `sampleListToProfile`

jlfwong (commented):
This is looking great! Thanks for following through with all the feedback. We're down to small nits. After those are addressed, this will be good to merge

src/import/trace-event.ts
```diff
@@ -365,50 +454,166 @@ function eventListToProfileGroup(events: TraceEvent[]): ProfileGroup {
...
+ */
+function frameInfoForSampleFrame({name, category}: StackFrame): FrameInfo {
+  return {
+    key: `${name}:${category}`,
+    name: name,
+  }
+}
+
+function getActiveFramesForSample(
+  stackFrames: TraceWithSamples['stackFrames'],
```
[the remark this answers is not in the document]
zacharyfmarion: It looks there are no references to `Record` which is a nice ts helper I use a lot - I will use `{[key: string]: StackFrame}` to keep consistency with the codebase.

src/import/trace-event.ts
zacharyfmarion: Yep, this is unnecessary

src/import/trace-event.ts
zacharyfmarion: Good point - I can just keep a counter and if the number of parents > total number of frames I can throw an error. Will hold off for now since it will make the code less readable but if you'd prefer I'm happy to make the change.

src/import/trace-event.ts
zacharyfmarion: Will add a type - there are four functions:

- `eventListToProfileGroup` which calls `constructProfileFromTraceEvents` for each thread
- `sampleListToProfileGroup` which calls `constructProfileFromSampleList` for each thread

Would you like me to rename to `eventListToProfile` and `sampleListToProfile` - IMO them being so similar in name makes them easy confuse.

zacharyfmarion:
@jlfwong awesome - attempted to address feedback with [this commit](https://github.com/jlfwong/speedscope/pull/453/commits/0a5633af0a416e2477c8a66389a8e7d5eb088a6e) and left a couple of questions

src/import/trace-event.ts
zacharyfmarion: Updated [here](c8c1e7f1af36f5fac5d62ee268ac62c56e3eda8b), seems sensible

jlfwong:
@zacharyfmarion Okay! This is deployed now on https://www.speedscope.app/

Are you still planning on following up with some hermes-specific changes? After that, would love to signal boost this since I'm sure there are other react native devs that could take advantage of these changes

zacharyfmarion:
@jfwong thank you so much! I will open a PR for the Hermes-specific changes today. I agree, once I'm back at work in the new year I can get my coworkers to help get the message out.

PR #458, opened 2023-12-26
Add hermes-specific support for the trace event format

zacharyfmarion:
Profiles that are transformed into the hermes trace event format are guaranteed to have specific arguments that supply metadata that is useful for debugging - namely the filename, line + col number that the function call originated from, with sourcemaps applied. This PR adds specific support for this information to the trace event importer. This means that we can have the Frame name be just the name of the function, since we know all the information we want to be displayed in the UI is captured in the frame info, which makes the traces cleaner to look at.

| Before | After |
|-----|-----|
| Screenshot 2023-12-26 at 2 40 01 PM | Screenshot 2023-12-26 at 2 39 13 PM |
| Screenshot 2023-12-26 at 2 41 03 PM | Screenshot 2023-12-26 at 2 41 29 PM |

### DIFF
--- a/src/import/trace-event.ts
+++ b/src/import/trace-event.ts
@@ -48,7 +48,7 @@ interface TraceEvent {
   cat?: string
 
   // Any arguments provided for the event. Some of the event types have required argument fields, otherwise, you can put any information you wish in here. The arguments are displayed in Trace Viewer when you view an event in the analysis section.
-  args: any
+  args?: any
 
   // A fixed color name to associate with the event. If provided, cname must be one of the names listed in trace-viewer's base color scheme's reserved color names list
   cname?: string
@@ -169,10 +169,7 @@ function selectQueueToTakeFromNext(
   // to ensure it opens before we try to close it.
   //
   // Otherwise, process the 'E' queue first.
-  const bFrontKey = keyForEvent(bFront)
-  const eFrontKey = keyForEvent(eFront)
-
-  return bFrontKey === eFrontKey ? 'B' : 'E'
+  return keyForEvent(bFront) === keyForEvent(eFront) ? 'B' : 'E'
 }
 
 function convertToEventQueues(events: ImportableTraceEvent[]): [BTraceEvent[], ETraceEvent[]] {
@@ -304,7 +301,10 @@ function keyForEvent(event: TraceEvent): string {
   return key
 }
 
-function frameInfoForEvent(event: TraceEvent, isHermesProfile: boolean): FrameInfo {
+function frameInfoForEvent<T extends boolean>(
+  event: T extends true ? HermesTraceEvent : TraceEvent,
+  isHermesProfile: T,
+): FrameInfo {
   const key = keyForEvent(event)
 
   // In Hermes profiles we have additional guaranteed metadata we can use to
@@ -313,10 +313,9 @@ function frameInfoForEvent(event: TraceEvent, isHermesProfile: boolean): FrameIn
     return {
       name: getEventName(event),
       key: key,
-      // Hermes specific
-      file: event?.args?.url,
-      line: event?.args?.line,
-      col: event?.args?.column,
+      file: event.args.url,
+      line: event.args.line,
+      col: event.args.column,
     }
   }
 
@@ -371,7 +370,7 @@ function getProfileNameByPidTid(
 function eventListToProfile(
   importableEvents: ImportableTraceEvent[],
   name: string,
-  isHermesProfile = false,
+  isHermesProfile: boolean = false,
 ): Profile {
   // The trace event format is hard to deal with because it specifically
   // allows events to be recorded out of order, *but* event ordering is still
--- a/src/import/trace-event.ts
+++ b/src/import/trace-event.ts
@@ -54,6 +54,24 @@ interface TraceEvent {
   cname?: string
 }
 
+interface HermesTraceEventArgs {
+  line: number | null
+  column: number | null
+  funcLine?: string | null
+  funcColumn?: string | null
+  name: string
+  category: string
+  parent?: number
+  url: string | null
+  params: string | null
+  allocatedCategory: string
+  allocatedName: string
+}
+
+type HermesTraceEvent = TraceEvent & {
+  args: HermesTraceEventArgs
+}
+
 interface BTraceEvent extends TraceEvent {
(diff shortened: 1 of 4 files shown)


src/import/trace-event.ts
```diff
@@ -48,12 +48,30 @@ interface TraceEvent {
   cat?: string
 
   // Any arguments provided for the event. Some of the event types have required argument fields, otherwise, you can put any information you wish in here. The arguments are displayed in Trace Viewer when you view an event in the analysis section.
-  args: any
+  args?: any
```
zacharyfmarion, line 51: Based on the spec this is not a required field for some event types, so marked it as optional. Happy to change back if this was intentional

src/import/trace-event.ts
```diff
@@ -641,6 +686,31 @@ function isTraceEventList(maybeEventList: any): maybeEventList is TraceEvent[] {
...
+  const requiredProperties: Array<keyof HermesTraceEventArgs> = [
+    'line',
+    'column',
+    'name',
+    'category',
+    'url',
+    'params',
+    'allocatedCategory',
+    'allocatedName',
+  ]
```
jlfwong, line 699: Browsers probably optimize this anyway, but just for sanity, let's move this array outside of the call-stack. This seems like it would fit well as a constant directly below `HermesTraceEventArgs` as `heremesTraceEventRequiredProperties`

jlfwong (commented):
Cool! High level looks good. A few minor requests inline, and then more generally please change all of the `isHermesProfile` arguments to be enum args instead of an `ExporterSource` with current values of `UNKNOWN` or `HERMES`. That way this will extend well if we want to specialize for other trace events types in the future, and also makes call-sites more self-documenting

src/import/trace-event.ts
jlfwong: Seems fine

src/import/trace-event.ts
```diff
@@ -669,6 +739,8 @@ export function importTraceEvents(rawProfile: Trace): ProfileGroup {
     return sampleListToProfileGroup(rawProfile)
   } else if (isTraceEventObject(rawProfile)) {
     return eventListToProfileGroup(rawProfile.traceEvents)
+  } else if (isHermesTraceEventList(rawProfile)) {
+    return eventListToProfileGroup(rawProfile, true)
```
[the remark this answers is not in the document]
zacharyfmarion: makes sense - generally I just make any function that takes in more than one obvious parameter take in an object so the meaning of every param is clear at the callsite, but I saw that it isn't a pattern in the codebase. Using an enum is super readable, will do that

zacharyfmarion:
@jlfwong thanks for the feedback, pushed up some fixes!

src/import/trace-event.ts
```diff
@@ -641,6 +705,23 @@ function isTraceEventList(maybeEventList: any): maybeEventList is TraceEvent[] {
...
+  return requiredHermesArguments.every(prop => prop in traceEventArgs)
+}
+
+function isHermesTraceEventList(maybeEventList: any): maybeEventList is HermesTraceEvent[] {
+  if (!isTraceEventList(maybeEventList)) return false
+
+  // We just check the first element to avoid iterating over all trace events,
+  // and asumme that if the first one is formatted like a hermes profile then
+  // all events will be
+  return isHermesTraceEvent(maybeEventList[0].args)
```
jlfwong, line 722: `[0]?.args` please!

jlfwong: Nevermind, this is guarded in `isTraceEventList`

jlfwong:
Amazing! One last change and then I think this should be all wrapped.

If you write a markdown description (just in a comment here is fine) for React Native profiling, then I'll copy that into the wiki and update the README to link to it to indicate there's specific support for React Native via Hermes.

Here's an example of such a page: https://github.com/jlfwong/speedscope/wiki/Importing-from-pprof-(go)

zacharyfmarion:
# Importing from Hermes

[Hermes](https://hermesengine.dev/) is a javascript engine developed by facebook for use in react-native applications. For the most up-to-date instructions on how to take a profile, see [Profiling with Hermes](https://reactnative.dev/docs/profile-hermes).

## Profiling in Development Mode

To record a sampling profiler from the Dev Menu:

1. Navigate to your running Metro server terminal.
2. Press d to open the Dev Menu.
3. Select Enable Sampling Profiler.

Screenshot 2023-12-27 at 8 51 15 AM

5. Execute your JavaScript by in your app (press buttons, etc.)
6. Open the Dev Menu by pressing d again.
7. Select Enable Sampling Profiler again (this is currently a bug and might correctly say disable in the future) to stop recording and save the sampling profiler.

A toast will show the location where the sampling profiler has been saved, usually in `/data/user/0/com.appName/cache/*.cpuprofile`

You can then extract the profile from your emulator / device using the following command:

```
npx react-native@latest profile-hermes [destinationDir]
```

To view, drag and drop the profile from destinationDir into Speedscope.

## Profiling in Release Mode

To profile hermes in a release build of your app, you can use the [react-native-release-profiler](https://github.com/margelo/react-native-release-profiler) npm package:

```
yarn add react-native-release-profiler
cd ios && pod install
```

1. Build your app in release mode
2. Start a profiling session:

```ts
import { startProfiling } from 'react-native-release-profiler'

startProfiling()
```

3. Stop the profiling session:

```ts
import { stopProfiling } from 'react-native-release-profiler'

// `true` to save the trace to the phone's downloads folder, `false` otherwise
const path = await stopProfiling(true)
```

4. Download and process the performance trace from your phone to your PC:

- On **Android**:
```
npx react-native-release-profiler --fromDownload --appId <your appId>
```
- On **iOS**:
```
npx react-native-release-profiler --local <path to profile>
```

To view, drag and drop the profile saved in your current working directory into Spedscope. It should be transformed with sourcemaps!

zacharyfmarion:
@jlfwong took a pass at some docs ^. Profiling in release mode is something that Margelo recently worked one while contracting for my company, so honestly these are probably the most up-to-date / complete docs that exist anywhere right now.

jlfwong:
@zacharyfmarion Great! Added to the wiki, and updated the README.

Whenever you'd like to do an awareness push on Twitter or LinkedIn or wherever, let me know and I'll help signal boost.

PR #460, opened 2024-01-03
Partition based on samples instead of traceEvents when importing a sample-based chrome trace

zacharyfmarion:
In the existing code, if `traceEvents` did not have any importable events, the profile would be marked as empty. This was a bug, as the dev Hermes profiles I was testing with had one X event which made the code work. However we do not need to guarantee (and the spec doesn't seem to) any traceEvents being present as long as we have samples and stack frames. 

When I tested a production profile taken from Hermes it did not have any importable events, just a metadata event with a thread name. This PR updates the implementation to iterate over partitioned samples instead of traceEvents so we construct profiles for all thread + process pairs referenced in the samples array.

| Before | After |
| --- | ----- |
| Screenshot 2024-01-03 at 9 58 56 AM | Screenshot 2024-01-03 at 9 58 17 AM |

### DIFF
--- a/sample/profiles/trace-event/simple-with-samples.json
+++ b/sample/profiles/trace-event/simple-with-samples.json
@@ -21,16 +21,6 @@
       "args": {
         "name": "mqt_js"
       }
-    },
-    {
-      "name": "mqt_js",
-      "cat": "mqt_js",
-      "ph": "X",
-      "dur": 0,
-      "pid": 7512,
-      "ts": "11550183666",
-      "tid": "7695",
-      "args": {}
     }
   ],
   "samples": [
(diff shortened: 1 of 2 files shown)


sample/profiles/trace-event/simple-with-samples.json
```diff
@@ -21,16 +21,6 @@
       "args": {
         "name": "mqt_js"
       }
-    },
-    {
```
zacharyfmarion, line 25: This confirms the fix - even without any importable events the profile still gets imported correctly.

jlfwong:
@zacharyfmarion Done!

zacharyfmarion:
@jfwong thank you!

Issue #465, opened 2024-03-01
Callgrind inputs: Some called functions along with their callees are detached from the rest of execution at the root level


glatosinski:
When visualizing the Callgrind output, some functions appear on the root level, while being called by other functions in the Callgrind file (they appear as `cfn` callees in other functions in the Callgrind output).

The example output for Callgrind ([download link](https://raw.githubusercontent.com/antmicro/openroad-callgrind-report/main/docs/_static/speedscope/callgrind/ariane/992185_5_1_grt_callgrind.out)):

![first-before-fix](https://github.com/jlfwong/speedscope/assets/46489721/47005ffc-3934-496f-88ce-c22a37ac4468)

Other visualizations demonstrating the issue:

![second-before-fix](https://github.com/jlfwong/speedscope/assets/46489721/090968b5-af6a-4279-9e7e-2c6fa78b46c0)

![third-before-fix](https://github.com/jlfwong/speedscope/assets/46489721/2aa6338a-f3b0-49a3-846c-4873b38c0cc8)

The expected behavior is to have those methods embedded in the main call graph, not separately, starting from the root level.

glatosinski:
I created a small PR (https://github.com/jlfwong/speedscope/pull/466) that fixes the issue.

PR #466, opened 2024-03-01
Callgrind parser: Fixed associating file names with symbols and detection of root nodes

glatosinski:
This PR introduces several small improvements to updating the current file name for the Callgrind parser:

* First of all, when called function (`cfn`) is within an inlined code block, it should use file names provided by `fi` or `fe` keys.
* Secondly, when `fn` key is present, the parser should set the current file name to the value provided in the previous `fl` key (such reset is also applied in KCacheGrind: [link to code](https://github.com/KDE/kcachegrind/blob/ea4314db2785cb8f279fe884ee7f82445642b692/libcore/cachegrindloader.cpp#L808).
* Thirdly, the file associated with `fn` should be stored separately so that it is not overwritten later by new `fl` calls.

Those small modifications improve detection of root nodes. Before the PR, the same function could have several files associated with it - when `fi`/`fe` are not taken into account, the file name for the current callee is taken from the last `fl` or `cfi` call, which is later added to `childMap` in `this.childrenTotalWeights` instead of the actual function with its file name. After this, only one of the two representations of the same symbol get deleted from `rootNodes`.

Without the improvement, we can observe detached call stacks for such "orphaned" functions. With the improvement, the functions have properly provided names, which leads to better generation of the call graph.

Some examples before and after applying the commit from the PR (the marked calls are for the same function):

* Example 1 ([callgrind file](https://raw.githubusercontent.com/antmicro/openroad-callgrind-report/main/docs/_static/speedscope/callgrind/ariane/992185_5_1_grt_callgrind.out))

  * Before:
    first-before-fix
  * After:
    first-after-fix

* Example 2

  * Before:
    second-before-fix

  * After:
    second-after-fix

* Example 3

  * Before:
    third-before-fix
  * After:
    third-after-fix

This fixes https://github.com/jlfwong/speedscope/issues/465


jlfwong:
Hi @glatosinski!

Thanks for contributing this, and sorry for the long response time. Can you please add a minimal test that demonstrates the failure mode this PR fixes?

Thanks!

glatosinski:
Hi, ok, I will prepare a minimal test

PR #484, opened 2024-11-01
Add support for Instruments 16 Time Profile Deep Copy

### LINKED ISSUE #481: XCode Instruments 16 breaks Deep Copy format import
Here's a sample Deep Copy from Instruments 16, which seems to have slightly different spacing:

```
Weight	Self Weight	Symbol Names
2.18 s  100.0%	0 s	 node (98674)
1.98 s  90.8%	0 s	  start
1.98 s   0.0%	0 s	   node::Start(int, char**)
1.98 s   0.0%	0 s	    node::LoadSnapshotDataAndRun(node::SnapshotData const**, node::InitializationResult const*)
1.98 s   0.0%	0 s	     node::NodeMainInstance::Run()
```

robert3005:
fix #481 

I generated profile using simple program, however, the capture profile is actually different than earlier instruments that's why there's new test snapshot


src/import/instruments.ts
```diff
@@ -38,12 +38,14 @@ interface PastedTimeProfileRow {
...
   'Source Path'?: string
   'Symbol Name'?: string
+  'Symbol Names'?: string
 }
 
 interface PastedAllocationsProfileRow {
   'Bytes Used'?: string
   'Source Path'?: string
   'Symbol Name'?: string
+  'Symbol Names'?: string
```
robert3005, line 48: This is still `Symbol Name` for allocations but if we don't add this field here we have a unknown field. Couldn't figure out nicer way of dealing with this

PR #493, opened 2025-01-14
Unbreak production builds in parcel

jlfwong:
Blocks #491

### DIFF
--- a/scripts/build-release.sh
+++ b/scripts/build-release.sh
@@ -27,4 +27,4 @@ node scripts/generate-file-format-schema-json.js > "$OUTDIR"/file-format-schema.
 cp assets/source-code-pro/LICENSE.md "$OUTDIR"/source-code-pro.LICENSE.md
 
 # Build the compiled assets
-node_modules/.bin/parcel build assets/index.html --no-cache --out-dir "$OUTDIR" --public-url "./" --detailed-report
+node_modules/.bin/parcel build assets/index.html --no-cache --dist-dir "$OUTDIR" --public-url "./" --detailed-report


jlfwong:
@cerisier Unfortunately the parcel upgrade, as it always does, introduces some changes that create a bunch of pain.

I can't land this yet because `scripts/build-release.sh` currently fails like so:

```
+ node_modules/.bin/parcel build assets/index.html --no-cache --dist-dir /Users/jlfwong/code/speedscope/dist/release --public-url ./ --detailed-report
🚨 Build failed.

@parcel/core: Failed to resolve 'preact' from './src/speedscope.tsx'

  /Users/jlfwong/code/speedscope/src/speedscope.tsx:1:25
  > 1 | import {h, render} from 'preact'
  >   |                         ^^^^^^^^
    2 | import {ApplicationContainer} from './views/application-container'
    3 | import {ThemeProvider} from './views/themes/theme'

@parcel/resolver-default: External dependency "preact" is not declared in package.json.

  /Users/jlfwong/code/speedscope/package.json:79:3
    78 |   },
  > 79 |   "dependencies": {
  >    |   ^^^^^^^^^^^^^^
    80 |     "open": "7.2.0"
    81 |   }
```
    
Accepting the suggestion would probably fix this bug, but it introduces an IMO unacceptable tradeoff -- it makes the `speedscope` npm package take on dependencies that are totally unneeded because it doesn't import them at all. This bloats what happens when people run `npm install speedscope`. The dependencies as listed are correct.
  
One solution for this is create a totally separate `package.json` for the CLI published through npm than I use for defining main project dependencies.

An alternative solution would be to complete the work I started in #295 and drop this annoying dependency entirely. I'm going to sleep on it and decide what to do later.

cerisier:
That's very unfortunate :( Especially since `parcel` is a bundler...
I wonder if `bundledDependencies` isn't made for this usecase, since you're bundling preact.
Maybe worth opening an issue on `parcel`.

jlfwong:
Abandoning in favor of #432

PR #496, opened 2025-01-23
Update README.md

dkuku:
Mention eflambe for erlang and elixir in the readme. The list is mostly sorted so I added it at the beginning.


jlfwong:
Hi! Happy to take this, but please add to the bottom of the list instead

dkuku:
The end is formatted a bit different than the rest. I hope it's ok now.

jlfwong:
Thanks!

PR #499, opened 2025-02-13
Update to working `open`, switch to ESM

flying-sheep:
See #498

I regenerated all the autogenerated files, but there was an issue with `demangle.wasm.js`: It *also* generated the `module.exports = Module; module.exports.default = Module` lines for some reason, so I manually deleted them.

If you know how to make it only emit ESM exports and leave out the CommonJS exports, I can add that to the makefile.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,18 +15,19 @@ jobs:
         node-version: [18.x, 20.x]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
       - run: npm test
         env:
           CI: true
-
+      - name: Try CLI
+        run: node bin/cli.mjs -v
       - name: Coveralls Parallel
-        uses: coverallsapp/github-action@master
+        uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.github_token }}
           flag-name: run-${{ matrix.node-version }}
@@ -37,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Coveralls Finished
-        uses: coverallsapp/github-action@master
+        uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.github_token }}
           parallel-finished: true
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -93,10 +93,10 @@ async function main() {
     fs.writeFileSync(jsPath, jsSource)
     urlToOpen += `#localProfilePath=${jsPath}`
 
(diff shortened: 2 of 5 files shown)


jlfwong:
Hey, unless I'm misunderstanding something important, updating to a version of `open` that works for you does not require all files be updated to esm — only the `bin/cli.js` script.

I'd prefer to keep this change as minimally scoped as possible.

Also, it would be helpful to understand the nature of the existing version of `open` not working for you. What errors are you seeing with the existing version of `open`?

flying-sheep:
It doesn’t support Plasma 6, so it’s completely useless and does nothing. Only version 10 is fixed.

flying-sheep:
OK, done. That requires renaming the CLI file, but of course that has much less impact than making everything ESM

bin/cli.mjs
```diff
@@ -90,11 +96,13 @@ async function main() {
...
-    fs.writeFileSync(htmlPath, `<script>window.location=${JSON.stringify(urlToOpen)}</script>`)
+    if (process.platform === 'darwin') {
+      const htmlPath = path.join(os.tmpdir(), `${filePrefix}.html`)
+      console.log(`Creating temp file ${htmlPath}`)
+      fs.writeFileSync(htmlPath, `<script>window.location=${JSON.stringify(urlToOpen)}</script>`)
 
-    urlToOpen = `file://${htmlPath}`
+      urlToOpen = `file://${htmlPath}`
+    }
   }
```
jlfwong, line 106: I've only actually tested this on OS X, so it's possible this issue exists for other platforms too.

To minimize the possibility of breakage, can we maintain this without the `process.platform` check? If you'd like to avoid this step on the specific platform you're using, I'm also open to skipping this step on any platforms you can test.

.github/workflows/ci.yml
```diff
@@ -15,18 +15,19 @@ jobs:
...
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
       - run: npm test
         env:
           CI: true
-
+      - name: Try CLI
+        run: node bin/cli.mjs -v
```
jlfwong, line 28: Thanks for adding this!

jlfwong:
Thanks for minimizing the surface area of this change! One more small tweak to that effect then this should be good to land

bin/cli.mjs
flying-sheep: Open supports 3 platforms: `win32`, `darwin`, and … `else`, so it tries to use `xdg-open` on all platforms that aren’t windows or macOS.

I tried on Windows, and indeed, the hash was also stripped, so I left it only for the platforms using `xdg-open`.

.github/workflows/ci.yml
flying-sheep: No problem!

flying-sheep:
OK, done! You were right about your platform concerns, I adapted my change.

jlfwong:
@flying-sheep Thanks! This is published on npm as part of speedscope@v1.22.1 and later

PR #502, opened 2025-03-16
Add support for FreeBSD's pmcstat callgraph format

ryan-moeller:
This eliminates the need for stackcollapse-pmc.pl and its dependency on Perl to import FreeBSD callgraphs.  Tracing can be done completely with tools included in the base system.  For example:

```
# kldload hwpmc
# pmcstat -S instructions -O sample.out sleep 1
# pmcstat -R sample.out -G sample.pmcstat.graph
```

To get `npm run jest` to run on FreeBSD, I changed the shebang in scripts/test-setup.sh from /bin/bash to /bin/sh as FreeBSD does not have /bin/bash. The script uses no bash-specific features. This is done in a separate commit.


src/import/index.ts
```diff
@@ -123,6 +124,9 @@ async function _importProfileGroup(dataSource: ProfileDataSource): Promise<Profi
   } else if (fileName.startsWith('callgrind.')) {
     console.log('Importing as Callgrind profile')
     return importFromCallgrind(contents, fileName)
+  } else if (fileName.endsWith('.graph')) {
```
jlfwong, line 127: I think, because this is a generic name, I'd rather omit this bit. The files should still import successfully because of the system below, it'll just be slightly slower as it hits tries a bunch of stuff that doesn't work. I'm nervous about this breaking imports for other profile types that just happen to have a .graph extension

If there's a more specific file pattern (like .pmc.graph or something) that works, I'd be happy to take that

ryan-moeller: The documentation for pmcstat uses .graph as the example file name so I used that, but I have no problem removing this extension matching shortcut.

Issue #468, opened 2024-03-07
Allow search by filename


gnodar01:
As an extension to #38, it would be great to be able to search by filename.

The problem I'm personally having is that I have an application which has a `Module` abstract class that requires all child classes to define a `run` function. The issue is that I have many such concrete classes that get run in a pipeline fashion together. So I'll have dozens of occurrences of the `run` function show up in the sandwich view (which I can search for), and when I hover any of them it will display the filename, e.g. `src/modules/someconcretemodule::run` and `src/modules/someotherconcretemodule::run`, but I can't search for "`src/modules/someconcretemodule`" specifically. Instead I have to go through the many occurrences of `run` until I find the one I'm looking for.

The general motivation is similar, it is often the case that the same function name appears many times throughout the codebase, but in different file. Being able to filter/search by function name AND filename, or just filename alone, would be very helpful.

Issue #246, opened 2019-12-04
FeatureRequest: only "my code"


guettli:
Feature Request: It would be nice if speedscope would have some kind of filtering.

Either white-listing: only include stacktrace lines which "mycode/mylib" in the path name

Or black-listing: exclude stacktrace lines which contain "framework/standard-library".

I am not able to implement this. But maybe the company I work for could sponsor/finance the implementation.

MidoriYakumo:
+1 Similar feature request for **ignore unknown sources and broken frames**

Proposals:
1. Add attributes **broken** for event at
https://github.com/jlfwong/speedscope/blob/16e32dc08e37948b931b6c9b132398f1c7493daf/src/import/linux-tools-perf.ts#L93
```javascript
event.sourceUnkown = event.stack.filter(({ symbolName: n }) => "[unknown]" !== n).length === 0
event.isIntrinsic = event.stack.filter(({ symbolName: n }) => "__" !== n.slice(0, 2)).length === 0
event.frameBroken = event.stack.filter(({ file: f }) => "[unknown]" === f).length > 0
```

2. Add feature switch on UI and toggle UI changes according these attributes
3. Time Order changes
original
![image](https://user-images.githubusercontent.com/6872220/103903708-33db6700-5137-11eb-9477-d4c042cb1534.png)
toggled: connect contiguous known events and gray out inaccurate broken samples
![image](https://user-images.githubusercontent.com/6872220/103904550-620d7680-5138-11eb-9bb2-9d994dc5294c.png)
toggled: detailed in timing info
![image](https://user-images.githubusercontent.com/6872220/103905010-08597c00-5139-11eb-966c-6a9130158a57.png)
4. Left Heavy changes
original
![image](https://user-images.githubusercontent.com/6872220/103905151-3474fd00-5139-11eb-8443-41bff2adc043.png)
toggled: gray out inaccurate and snapped samples
![image](https://user-images.githubusercontent.com/6872220/103905781-fdebb200-5139-11eb-8167-ab09a6483843.png)
toggled: detailed in timing info similar with Time Order

Issue #192, opened 2018-11-15
Allow to filter graph


Yajo:
It would be awesome to be able to filter in and out visible results.

For instance, if I develop a module of a given framework, I want to exclude the framework files; but maybe if I develop the framework, I want to exclude the module files.

PR #503, opened 2025-03-16
Add functionality to filter flame graphs by search query

vivster7:
Adds a 🎯 button to the ctrl+f search bar that toggles an 'exact match' filter mode -- it only shows frames that exactly match the search query.

https://github.com/user-attachments/assets/43b0d5ea-373f-41c9-a463-348998eacc91

This is one possible implementation for https://github.com/jlfwong/speedscope/issues/192

The use case I was trying to support (for myself), was the ability to simplify large call graphs into first party code vs specific third party libraries. Seems to kind of relate to the following issues as well
- https://github.com/jlfwong/speedscope/issues/246
- https://github.com/jlfwong/speedscope/issues/468

Filtering the profile on keystroke feels like a fairly expensive operation, but I wasn't sure how else to implement. My fork solves my use case, but would be nice upstream if you'd accept this patch or a similar variant.

### DIFF
--- a/src/app-state/index.ts
+++ b/src/app-state/index.ts
@@ -14,6 +14,9 @@ export const flattenRecursionAtom = new Atom<boolean>(false, 'flattenRecursion')
 export const searchIsActiveAtom = new Atom<boolean>(false, 'searchIsActive')
 export const searchQueryAtom = new Atom<string>('', 'searchQueryAtom')
 
+// True if the flamechart should only show the matching frames when searching.
+export const onlyMatchesAtom = new Atom<boolean>(false, 'onlyMatches')
+
 // Which top-level view should be displayed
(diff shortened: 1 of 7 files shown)


jack-mcivor:
@vivster7 does this implement searching by filename #468 ? 

I think that for my profile (Python code, generated with Autin) filtering by filename is the only way to achieve a distinction between first party code & specific third party libraries

vivster7:
@jack-mcivor I think so -- it matches against the frame's name, so if your frame contains the filename, you should be able to filter the "time order" and "left heavy" views by filename. I don't think this will do anything to the sandwich view though (which is maybe specifically what https://github.com/jlfwong/speedscope/issues/468 was asking)

PR #527, opened 2025-10-21
chrome: tolerate missing args data

jefflee-figma:
This change allows Speedscope to tolerate Chrome profile chunks where `args.data` is undefined. It's unclear whether this is a Chrome bug, but we are noticing recent profiles that crash Speedscope without this tweak. FWIW, it also appears that Firefox seems to die on a similar issue.

Happy to provide an example trace that repros, but out of an abundance of caution I'd prefer to email rather than post publicly. Feel free to shoot me an email: jefflee@figma.com


jlfwong:
@jefflee-figma Okay, this seems pretty harmless to me

Can you add a minimal test to this (file should be as small as possible) to prevent this from regressing in the future?

jefflee-figma:
> can you move the test file into the appropriate directory in samples/profiles/Chrome/{whatever version}, and similarly move the test to chrome.test.ts

Yep, thanks for clarifying that. This is addressed in ac3b0c76594d5e389150cc4bb2d7fd46f88befbb. I assumed that the directory name should be the major version listed under "About Chrome", but let me know if that is supposed to be something else.

Issue #244, opened 2019-12-02
No obvious error if WebGL is disabled


baldurk:
If WebGL is disabled in the browser the site loads and the browse button works, but nothing happens after selecting a file. Looking in the console there are these messages:

```
Error: Setup failure - speedscope.f741b731.js:93:3612
Error: Cannot find module '109' - demangle-cpp.6caf93ee.js:1:302
Error: Cannot find module '101' - demangle-cpp.6caf93ee.js:1:302
```

I looked up the mangled code and that error does seem to be thrown if `getContext("webgl", ...)` fails, so it seems to check out. I don't know how but some other sites will show a more 'friendly' error if WebGL is disabled but the site requires it.

OS: Windows 7
Browser: firefox 70.0.1 (64-bit) with `webgl.disabled` set to `true` in `about:config`
I tried chrome and it worked fine, I couldn't see how to disable webgl on chrome (there's no setting at least) so I couldn't test if that would cause it to fail on chrome too.

jlfwong:
Ah, yeah, I don't think I ever tested this. I'd be happy to accept a PR which makes puts out a warning that's clearer about this. I think a simple `alert()` would suffice for now.

PR #531, opened 2025-11-25
Add error message when WebGL setup fails

cmdcolin:
I am on Ubuntu 25.10 and for whatever reason,  the normal Google Chrome .deb file that you have to download and install manually does not have webgl enabled (on my machine only? who knows)

The speedscope app doesn't give any user visible error in this case though, so I made it show the default error atom here

image

Fixes https://github.com/jlfwong/speedscope/issues/244

BTW huge thank you for speedscope, absolutely great app for performance profiling


jlfwong:
Thanks!

Issue #397, opened 2022-06-20
Java Flight Recorder support


vasi:
Would y'all be interested if I implemented JFR support? There's not really any official format documentation, but there's [some basic docs here](https://www.morling.dev/blog/jdk-flight-recorder-file-format/) and an [Apache licensed Java parser](https://github.com/jvm-profiling-tools/async-profiler/blob/master/src/converter/one/jfr/JfrReader.java) to steal from.

Unfortunately it's a binary format, but there's good support for [binary reads](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) in JS nowadays, so it ought to be possible.

parttimenerd:
Reading the JFR format in JS is considerable effort and possibly leads to maintenance issues.

parttimenerd:
A converter with basic support can be found here: https://github.com/parttimenerd/jfrtofp

@jlfwong is there the possibility to add this to your converter pages?

jlfwong:
This looks reasonable to me, with two changes:

1. Please make the esbuild upgrade into a separate PR, and let's get that landed first
2. The wasm ball is going to need to be lazy loaded so we don't block initial load on the download of the wasm ball for this

jlfwong:
Going to close this as done w/ #533

PR #533, opened 2025-11-30
Basic Java Flight Recorder support

beatbrot:
As promised in #397: JFR file support.

I did not add tests since the test setup for WASM files is just too cumbersome. Hope this is okay :)


jlfwong (commented):
Thanks for this @beatbrot!

Can you say more about what makes the test setup cumbersome here? I'm generally reluctant to have format support that is just wholly untested.

jlfwong:
@beatbrot This is now live on https://speedscope.app and published to npm as speedscope@1.25.0

Two additional things...
1. Can you please write a markdown file like the one here explaining how to use this? https://github.com/jlfwong/speedscope/wiki/Importing-from-async%E2%80%90profiler-(Java)
2. The downside of this using fetch to grab the wasm data is the JFR import won't work when speedscope is used as static files w/ no server locally. This is a workflow we've historically supported, as described here: https://github.com/jlfwong/speedscope?tab=readme-ov-file#self-contained-directory. It seems like browsers are hellbent on kneecapping support for features that aren't running under https (including not running under file:// protocol), so perhaps this is fine. If someone hits this, they can report it and we can figure out how to address

jlfwong:
Thanks! Just updated here: https://github.com/jlfwong/speedscope?tab=readme-ov-file#supported-file-formats

> Hmmm okay... to be honest, I have no idea how we could solve this properly though...

I think for now we just ignore it and see if anyone cares. If they do, this is, I believe, fixable by base64 encoding the wasm binary and putting it into a lazy-loaded `.js` script so it doesn't need to be fetched from a `file://` protocol

PR #542, opened 2026-02-17
feat: add --serve flag to speedscope to run basic webserver

### LINKED ISSUE #541: feat: support webserver for speedscope cli
I'm working with callgrind files in a devcontainer. Unfortunately, this means that the "open in browser" behavior native to speedscope doesn't quite work "out of the box". 

If possible, I'd love to see a little  webserver built into speedscope so that I can run something like:

```
speedscope cachegrind.out.38776.gz --serve
Serving speedscope on http://localhost:46387
```

so that I can just quickly visualize the callgrind file. 

Optionally, and I don't know how big an ask this is, it would be nice to extract all the various `Parser` -> `ProfileGroup`  behaviors out of the react app into a separate lib so that I could potentially write this little server myself. In its current state, `speedscope` encapsulates the entire react app onto npm making it difficult to extend.

Aside, this project is awesome.

damienwebdev:
Closes #541


src/views/application.tsx
```diff
@@ -568,6 +569,9 @@ export class Application extends StatelessComponent<ApplicationProps> {
     }
 
     if (!activeProfileState || !glCanvas) {
+      if (this.props.hashParams.localProfilePath || this.props.hashParams.profileURL) {
```
damienwebdev, line 572: I added this to avoid a flicker of the "browse" page when the app is loading an existing file with `--serve`

PR #551, opened 2026-05-13
fix: Safari 26.4 timeline export support

agurtovoy:
First of all, thanks for the great tool, super helpful!

At some point Safari changed its export format without bumping up the version. This PR adds support for Safari 26.4 timeline export while remaining backward compatible with the older format.


src/import/safari.ts
```diff
@@ -48,9 +53,15 @@ interface Recording {
...
-  sampleStackTraces: Sample[]
-  sampleDurations: number[]
-}
+} & (
+  | {
+      sampleStackTraces: StackFramesSample[]
+      sampleDurations: number[]
+    }
+  | {
+      samples: Sample[]
```
agurtovoy, line 62: The new format moves `sampleStackTraces` and `sampleDurations` into the `samples` array (e.g. `samples: [{ stackTraces: [], durations: []}]`)

PR #552, opened 2026-05-22
Fix callgrind parser: positions: instr / positions: instr line and cyclic call graphs

marc-casavant:
Fixes Issue #550

This is a larger-than-usual PR, but the changes are tightly coupled. Each is required to successfully import callgrind.out.* files produced when profiling FreeRADIUS with Valgrind. Splitting them apart would leave the parser in a broken intermediate state.

### Background
The callgrind format allows a `positions:` header that declares how many position columns precede cost values on each cost line. The most common value is `positions: line` (one column), but profilers targeting native code routinely emit `positions: instr line` (two columns) or `positions: instr` (one column).  One example of this would be Valgrind profiles of FreeRADIUS. 

The parser previously hard-coded `numPositionFields = 1`; any file with a non-default `positions:` header would silently produce garbage output. Separately, call graphs that are entirely cyclic (every function has at least one caller) had no natural root nodes and produced an empty flame graph.

The following changes have been made to allow Speedscope to display flame graphs of Valgrind profiles generated by running tests on FreeRADIUS.

### Changes
**`positions: `header support**

- Parse the `positions:` header and store `numPositionFields` (count of whitespace-separated tokens).
- Pass `numPositionFields` into `parseCostLine` so the correct number of leading columns are skipped when extracting cost values.
- Add hex address parsing (`0x…` tokens) so instruction-address position fields are consumed correctly and subposition compression still works on subsequent lines.

**Cyclic call graph fallback**

- After the natural-roots pass, if `rootNodes` is empty, fall back to a residual-weight strategy: compute each frame's total incoming edge weight, subtract it from its total weight, and use frames with positive residual as synthetic roots, visited heaviest-first.
- Add a `visitedGlobally` set to avoid exponential re-expansion of shared nodes reachable from multiple roots.
- Emit a `console.warn` when cycles are detected.

### New tests

- `importFromCallgrind positions instr line` 
  - two-column `positions: instr line`; instruction address skipped, costs parsed correctly
- `importFromCallgrind positions instr`
  - one-column `positions: instr`; hex address consumed, no cost corruption
- `importFromCallgrind instr line subposition compression`
  - subposition compression (`*`, `+N`, `-N`) works correctly across two position columns
- `importFromCallgrind cyclic callgraph`
  -  fully-cyclic graph produces a non-empty flame graph via residual-weight roots

### Reference file
A callgrind.out.* file (1.3 MB, generated by Valgrind while profiling FreeRADIUS) is attached as an example of the type of files we are viewing with Speedscope.

[callgrind.out.20-11.txt](https://github.com/user-attachments/files/28157444/callgrind.out.20-11.txt)

### DIFF
--- a/sample/profiles/callgrind/callgrind.cyclic-callgraph.log
+++ b/sample/profiles/callgrind/callgrind.cyclic-callgraph.log
@@ -1,0 +1,20 @@
+# callgrind format
+events: Ir
+
+# funcA calls funcB, funcB calls funcA — fully cyclic, no natural root nodes.
+# The importer must fall back to residual-weight root finding (option 2).
+# funcA residual = 400 - 150 = 250  (positive → root)
+# funcB residual = 200 - 300 = -100 (negative → not a root)
+
+fl=main.c
+fn=funcA
+1 100
+cfn=funcB
+calls=1 10
+10 300
+
+fn=funcB
+10 50
+cfn=funcA
+calls=1 1
+1 150
--- a/sample/profiles/callgrind/callgrind.instr-line-positions.log
+++ b/sample/profiles/callgrind/callgrind.instr-line-positions.log
@@ -1,0 +1,33 @@
+# callgrind format
+positions: instr line
+events: Instructions
+
+fl=file1.c
+fn=main
+0x08048001 16 20
+jfi=file1.c
+jfn=main
+jcnd=2/5 0x08048005 16
+0x08048005 16 3
+jfl=file1.c
+jfn=main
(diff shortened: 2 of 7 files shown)


jlfwong:
Hi @marc-casavant, thanks for the patch and the explanation! I basically never use callgrind profiles, so I'm taking it mostly on faith that this patch is good, and that if bugs come up, I'll tag you in them to look at.

AI review surfaced one potential issue. Would you mind taking a look and seeing if it's a real problem, or just noise? Happy to merge this after

